Issue #10, opened 2026-04-30
Make installation tractable for non-technical users (Native Messaging + installers)


ylohnitram:
## Goal

Right now the install flow assumes the user can: install Node, run `npm install -g`, run a `claude login` interactive flow, clone the repo, keep `node score-server.js` running, manage a localhost port, copy-paste an auth token, and side-load an unpacked extension. Non-technical users (e.g. recruiters using their LinkedIn for sourcing) cannot do this.

The goal is **5 minutes from "I want to try this" to "it's running"** without the user typing a single shell command after the installer dialog. Default backend stays **flat-rate Claude Code subscription** — no Anthropic API tokens.

## Architecture change

Replace the localhost HTTP bridge (`score-server.js` + auth token) with a **Chrome Native Messaging host**.

```
Chrome MV3 extension (linkshit)
└─ background.js → chrome.runtime.connectNative('com.replikanti.linkshit')
                    ↓
                  native messaging host  (~/.linkshit/host.js)
                    ↓
                  spawns `claude -p`
                    ↓
                  user's Pro/Max subscription (flat)
```

Why this is better:
- Chrome spawns the host on demand and tears it down when the connection closes — no server process to manage.
- No localhost port — no firewall, no port collisions.
- No auth token — Chrome authenticates connections via `allowed_origins` in the native host manifest, keyed by extension ID (which is stabilized via the `key` field in `manifest.json`).
- Settings UI shrinks — no Server URL, no Server Token. Just Criteria, filters, threshold.

## What goes in `advanced/`

The Anthropic API key path remains as opt-in for users who don't have a Claude Pro/Max subscription. Same `runLLM()` swap point as today, just now in `host.js`. Documented in README under "Advanced — bring your own API key".

The legacy localhost HTTP bridge stays as a documented fallback for environments where Native Messaging is blocked by enterprise Chrome policy.

## Sub-PRs

1. **PR-A — Native Messaging host (this issue's core)**
   - `key` field in `manifest.json` for stable extension ID
   - `permissions: ["nativeMessaging"]`
   - `background.js` service worker — bridges `chrome.runtime.sendMessage` from content script to `chrome.runtime.connectNative` to the host
   - `host.js` — Node native messaging host, length-prefixed stdin/stdout JSON, `runLLM()` swap point spawns `claude -p` by default
   - `content.js` — drop server URL / server token from settings UI; route batches through `chrome.runtime.sendMessage`
[...]

PR #3, opened 2026-04-29
ci: add primary GitHub Actions workflow (lint / check / pack)

ylohnitram:
## Summary

Adds the primary CI workflow at `.github/workflows/ci.yml`, plus the dev tooling it drives. Runs on every push to `main` and every PR.

Pipeline stages:

| Step | Command | What it catches |
|---|---|---|
| Install | `npm ci` | Lockfile drift |
| Validate | `npm run validate` | Broken JSON in `manifest.json` or `package.json` |
| Check | `npm run check` | JS syntax errors in `content.js` and `score-server.js` |
| Lint | `npm run lint` | ESLint flat-config — browser globals for content.js, node globals for score-server.js |
| Pack | `npm run pack` | `web-ext build` produces a Chrome-loadable ZIP |
| Upload | `actions/upload-artifact@v4` | 14-day retention, `if-no-files-found: error` |

Verified locally: lint clean, `web-ext build` produces a 6.3 KB ZIP containing exactly `manifest.json` + `content.js` (verified via `unzip -l`).

## Files

- `.github/workflows/ci.yml` — workflow definition (Node 20, Ubuntu)
- `eslint.config.mjs` — flat config (no preset framework, just `@eslint/js` recommended + per-file globals)
- `package.json` — adds `validate` / `check` / `lint` / `pack` scripts and dev dependencies (`eslint`, `@eslint/js`, `globals`, `web-ext`)
- `package-lock.json` — committed for reproducible `npm ci`
- `.gitignore` — ignores `web-ext-artifacts/`
- `README.md` — CI badge + Development section

## Out of scope (deferred to follow-up PR)

The user requested a separate workflow with additional checks appropriate for a public browser extension. That will land as `extras.yml` in a follow-up PR and may include:

- `npm audit` for known CVEs
- `web-ext lint` (Mozilla's stricter manifest linter)
- CodeQL static analysis
- README markdown link check
- Smoke test of `score-server.js` (boot, exercise auth gate, kill)

## Test plan

- [ ] CI workflow appears under Actions tab and runs on this PR
- [ ] All five steps (validate, check, lint, pack, upload) green
- [ ] `linkshit-extension` artifact is downloadable from the workflow run
- [ ] Downloaded ZIP loads as an unpacked extension in Chrome with no manifest errors
- [ ] CI badge in README renders and links to the workflow page

PR #12, opened 2026-04-30
feat: non-tech-friendly installer + GitHub Releases pipeline (refs #10)

ylohnitram:
## Summary

PR-B from [#10](https://github.com/Replikanti/linkshit/issues/10). After this lands and v0.2.0 is tagged, a non-technical user gets from "I want this" to "it works" by:

1. Downloading **`install.sh`** from the Releases page → running `bash install.sh`
2. Downloading **`linkshit-extension-vX.Y.Z.zip`** → unzipping → drag into `chrome://extensions` (Developer mode on)
3. Setting Criteria in the panel and clicking Start.

No `git clone`, no `npm install`, no manual native-messaging-manifest writing, no terminal command beyond `bash install.sh`. Linux + macOS only — Windows is intentionally out of scope.

### `installers/install.sh`

Single bash script. Installs Node-detected, Claude Code globally, drops `host.js` into `~/.linkshit/`, writes the Chrome native messaging manifest at the OS-specific path, prints the rest of the steps. Two ways to obtain `host.js`:

- **Released form** (`install.sh` from the Releases page): host.js is embedded inline as base64 in the script. No network fetch at install time.
- **Template form** (running from a clone): falls back to downloading host.js from the latest GitHub release.

### `.github/workflows/release.yml`

Tag-triggered (`v*`). Builds the extension zip, embeds host.js into `install.sh` (with round-trip verification), creates a GitHub Release via `gh release create`, attaches:

- `linkshit-extension-vX.Y.Z.zip`
- `host.js`
- `install.sh`

### `.github/workflows/extras.yml`

New `shellcheck` job — runs `shellcheck installers/install.sh` on every push and PR.

### `docs/screenshots/`

Three SVG mockups embedded in README:

| | |
|---|---|
| `01-extensions-dev-mode.svg` | chrome://extensions with Developer mode toggle highlighted |
| `02-load-unpacked.svg` | Same page with Load unpacked button + folder picker |
| `03-panel-on-linkedin.svg` | The Linkshit panel on a LinkedIn feed page |

Mockups, not real screenshots. README explicitly notes this and invites real PNGs as a follow-up.

### `README.md`

New **Quick Start** section at the top (3 steps, ~5 min, embeds the SVG mockups inline). Old developer setup demoted to **Manual setup (developer-grade)**. Files list and troubleshooting updated.

## Verified locally

- `npm run lint` / `check` / `validate` / `pack` — green.
- `shellcheck installers/install.sh` — clean.
- Python `yaml.safe_load` on `release.yml` — valid.
- **Embed simulation:** ran the same `sed` substitution the workflow does, parsed result with `bash -n`, asserted placeholder is gone, decoded the embedded base64, and `diff -q`'d against `host.js` — byte-for-byte match.

## Out of scope (queued for PR-C)

- Native-host smoke test in CI. The smoke job was removed in PR-A together with `score-server.js`; PR-C re-adds it pointed at `host.js` (driving the Native Messaging stdin/stdout protocol).

## Test plan

- [ ] CI green (ci, web-ext lint, npm audit, link check, shellcheck, CodeQL, SonarCloud)
- [ ] Tag v0.2.0 after merge → release workflow triggers → assets appear on the Releases page
- [ ] Reviewer downloads `install.sh` from the v0.2.0 release, runs it on a fresh Linux/macOS box, confirms the panel works on LinkedIn

### DIFF
--- a/.github/workflows/extras.yml
+++ b/.github/workflows/extras.yml
@@ -49,6 +49,14 @@ jobs:
         # Chrome-targeted MV3 extension. Real ERRORS still fail.
         run: npx --no-install web-ext lint --source-dir dist --self-hosted
 
+  shellcheck:
+    name: shellcheck installer
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: ShellCheck install.sh
+        run: shellcheck installers/install.sh
+
   links:
     name: Markdown link check
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,24 +42,40 @@ jobs:
           # Encode host.js as base64 (single line, no wrap) and substitute
           # into the installer template. Use `|` as the sed delimiter since
           # the base64 alphabet (A-Za-z0-9+/=) does not include it.
+          #
+          # The sed is anchored to the assignment line (`^readonly
+          # LINKSHIT_HOST_JS_BASE64=`) so it only rewrites that one place
+          # and never accidentally touches another occurrence of the
+          # placeholder elsewhere in the script (would-be bug: a string
+          # comparison line would be substituted too and silently neutralise
+          # the embedding, dragging every released installer back onto the
+          # network-fallback branch).
           HOST_B64=$(base64 -w0 host.js)
-          sed "s|__HOST_JS_BASE64__|${HOST_B64}|" \
+          sed "/^readonly LINKSHIT_HOST_JS_BASE64=/s|__HOST_JS_BASE64__|${HOST_B64}|" \
             installers/install.sh > installers/install.release.sh
           chmod +x installers/install.release.sh
-          # Sanity-check: parse the generated script.
+
+          # Verify: parses, placeholder gone from the assignment line, blob
+          # round-trips byte-for-byte, and the embedded blob is long enough
+          # that install.sh's runtime length-based detection takes the
+          # embedded branch (not the network fallback).
           bash -n installers/install.release.sh
-          # And spot-check that the placeholder is gone and the encoded
-          # blob round-trips back to the original host.js.
-          if grep -q '__HOST_JS_BASE64__' installers/install.release.sh; then
-            echo "ERROR: placeholder still present after substitution" >&2
+          if grep -q '^readonly LINKSHIT_HOST_JS_BASE64="__HOST_JS_BASE64__"$' installers/install.release.sh; then
+            echo "ERROR: assignment-line placeholder not substituted" >&2
             exit 1
           fi
-          ENCODED=$(grep -oP 'LINKSHIT_HOST_JS_BASE64="\K[^"]+' installers/install.release.sh)
+          ENCODED=$(grep -oP '^readonly LINKSHIT_HOST_JS_BASE64="\K[^"]+' installers/install.release.sh)
           echo "$ENCODED" | base64 -d | diff -q - host.js || {
             echo "ERROR: embedded host.js does not round-trip to source" >&2
             exit 1
           }
-          echo "✓ install.release.sh built and verified"
+          if [ "${#ENCODED}" -lt 200 ]; then
+            echo "ERROR: embedded base64 is only ${#ENCODED} chars; install.sh's" >&2
+            echo "       length-based template detection (<200) would route this" >&2
+            echo "       installer to the network-fallback path." >&2
+            exit 1
+          fi
+          echo "✓ install.release.sh built and verified (${#ENCODED} chars embedded)"
 
       - name: Create GitHub Release with assets
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,73 @@
+name: Release
+
+# Triggered when a tag matching v* is pushed. Builds the extension zip,
+# embeds host.js into the installer template, and publishes everything as
+# attached assets on a GitHub Release.
+#
+# Assets shipped on each release:
+#   - linkshit-extension-vX.Y.Z.zip   (Chrome MV3 unpacked extension)
+#   - host.js                         (raw, for advanced users)
+#   - install.sh                      (embedded host.js inline; user runs this)
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 22
+          cache: npm
+
+      - run: npm ci
+
+      - name: Pack the extension
+        run: |
+          npm run pack
+          # Rename to include the tag for clarity in the Releases UI.
+          mv "$(ls web-ext-artifacts/linkshit-*.zip)" \
+             "web-ext-artifacts/linkshit-extension-${GITHUB_REF_NAME}.zip"
+
+      - name: Embed host.js into install.sh
+        run: |
+          # Encode host.js as base64 (single line, no wrap) and substitute
+          # into the installer template. Use `|` as the sed delimiter since
+          # the base64 alphabet (A-Za-z0-9+/=) does not include it.
+          HOST_B64=$(base64 -w0 host.js)
+          sed "s|__HOST_JS_BASE64__|${HOST_B64}|" \
+            installers/install.sh > installers/install.release.sh
+          chmod +x installers/install.release.sh
+          # Sanity-check: parse the generated script.
+          bash -n installers/install.release.sh
+          # And spot-check that the placeholder is gone and the encoded
+          # blob round-trips back to the original host.js.
+          if grep -q '__HOST_JS_BASE64__' installers/install.release.sh; then
+            echo "ERROR: placeholder still present after substitution" >&2
+            exit 1
+          fi
+          ENCODED=$(grep -oP 'LINKSHIT_HOST_JS_BASE64="\K[^"]+' installers/install.release.sh)
+          echo "$ENCODED" | base64 -d | diff -q - host.js || {
+            echo "ERROR: embedded host.js does not round-trip to source" >&2
+            exit 1
+          }
+          echo "✓ install.release.sh built and verified"
+
+      - name: Create GitHub Release with assets
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release create "${GITHUB_REF_NAME}" \
+            --title "${GITHUB_REF_NAME}" \
+            --generate-notes \
+            "web-ext-artifacts/linkshit-extension-${GITHUB_REF_NAME}.zip" \
+            "host.js" \
+            "installers/install.release.sh#install.sh"
--- a/README.md
+++ b/README.md
@@ -14,10 +14,56 @@ Claude Pro/Max subscription (no API tokens, no per-call cost). Replace one
 function in `host.js` and you can point it at the Anthropic API, Ollama, or
 anything else.
 
-> **Heads-up.** A non-technical one-click installer is in the oven (issue
-> [#10](https://github.com/Replikanti/linkshit/issues/10), shipping with
-> v0.2.0). Until then the setup below is developer-grade — you write the
-> Chrome native messaging manifest once by hand.
+## Quick Start (5 minutes, no terminal commands to memorize)
+
+> Linux and macOS only. Windows is intentionally not supported.
+
+### 1. Run the installer
+
+Download `install.sh` from the [latest release](https://github.com/Replikanti/linkshit/releases/latest), then in a terminal:
+
+```bash
+bash ~/Downloads/install.sh
+```
+
+The installer will:
+
+1. Verify Node.js 22+ is installed (and tell you where to get it if not).
+2. Install the `claude` CLI globally (Claude Code; ships your Pro/Max OAuth login).
+3. Drop `host.js` into `~/.linkshit/`.
+4. Register Chrome's native messaging manifest in the right OS-specific path.
+
+If the script tells you to log in to Claude Code first, open a new terminal,
+run `claude`, finish the sign-in flow in your browser, type `/exit`, then
+re-run `install.sh`.
+
+### 2. Load the extension in Chrome
+
+Download `linkshit-extension-vX.Y.Z.zip` from the same release, unzip it
+somewhere stable (e.g. `~/Linkshit/`).
+
+Open `chrome://extensions` and turn **Developer mode** on (top right):
+
+![Step 1 — turn Developer mode on](docs/screenshots/01-extensions-dev-mode.svg)
+
+Then click **Load unpacked** and pick the unzipped folder:
+
+![Step 2 — Load unpacked, select folder](docs/screenshots/02-load-unpacked.svg)
+
+### 3. Use it
+
+Visit [`linkedin.com/feed`](https://www.linkedin.com/feed). A panel appears
+top-right. Click ⚙, set your **Criteria** (one or two sentences describing
+what's worth surfacing), Save, then **Start**.
+
+![The Linkshit panel on a LinkedIn feed page](docs/screenshots/03-panel-on-linkedin.svg)
+
+Hits stream in sorted by score. Click **Open on LinkedIn →** on any of them
+to jump to the post.
+
+> **Note on the screenshots above.** They are mockups, not real Chrome
+> captures — Linkshit's CI does not run a browser. Replace the SVGs in
+> `docs/screenshots/` with real PNGs whenever convenient.
 
 ## How it works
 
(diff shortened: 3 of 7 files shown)


ylohnitram:
## QA Report

> **Update (2026-04-30):** all 3 actionable findings (#1 HIGH embed bug, #2 MED `read` EOF, #3 LOW workflow assertion gap) addressed in [`882f86f`](https://github.com/Replikanti/linkshit/pull/12/commits/882f86f). #4 INFO is a nit, no action. Two-layer fix on #1: install.sh detects template-vs-released by length (not string equality, so sed can't mistakenly substitute the comparison literal) AND release.yml's sed is anchored to the assignment line (defense in depth). Workflow now also asserts the embedded blob is ≥200 chars so the same class of bug fails the build loudly. End-to-end verified locally: released form routes to EMBEDDED branch (7672 chars), template form routes to TEMPLATE/network branch (18 chars), `bash install.sh </dev/null` exits 0 with the friendly "log in first" message. Original verdict preserved below for the audit trail.

### Verdict
❌ Block

### Summary
The installer, release pipeline, README, and SVG mockups are all sound on the surface — shellcheck clean, YAML parses, the embed round-trip passes, the extension still builds. **However, the substitution that embeds `host.js` into `install.sh` rewrites two lines instead of one, which silently neutralises the embedding and forces every released installer to download `host.js` from the network on every run.** The workflow's own sanity checks pass anyway, so the bug ships invisibly. There is also a smaller UX issue with `set -euo pipefail` + `read` that fails the script with no message if stdin is closed.

### Findings

| 1 | HIGH | release pipeline / installer | `sed "s\|__HOST_JS_BASE64__\|...\|"` in `release.yml` substitutes BOTH occurrences of the placeholder — the assignment on line 32 and the **comparison literal on line 117**. The released installer then compares the embedded blob to itself, the test is always true, and execution falls into the network-download branch. The embedded base64 is never decoded. | ✅ fixed in 882f86f |
| 2 | MEDIUM | installer UX | `set -euo pipefail` + `read -r -p ...` exits the script with code 1 and zero output if the user closes stdin (Ctrl-D, or any non-TTY invocation like `bash install.sh </dev/null`). User has no idea what happened. | ✅ fixed in 882f86f |
| 3 | LOW | release.yml | Workflow's "spot-check" only verifies that the placeholder is gone and that the encoded blob round-trips — neither check would catch finding #1, because both checks are satisfied while the conditional has been silently broken. Suggest also asserting the post-substitution script's runtime branch (e.g. dry-run with a fake `claude` and `node` and confirm no network fetch is attempted). | ✅ fixed in 882f86f |
| 4 | INFO | release.yml | Workflow uses unbounded glob `ls web-ext-artifacts/linkshit-*.zip` for the rename. Verified the produced filename is `linkshit-0.2.0.zip` so the glob matches one file — fine for now, but if `npm run pack` ever emits a sourcemap or sidecar the rename would break. | ✅ |

### Detailed verification

**1. shellcheck and bash parse — clean.**
```
$ shellcheck installers/install.sh && echo PASS
PASS
$ bash -n installers/install.sh && echo PASS
PASS
```

**2. YAML files parse.**
```
$ python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"
$ python3 -c "import yaml; yaml.safe_load(open('.github/workflows/extras.yml'))"
```

**3. Embed simulation (the four checks the spec requires) — all pass on their own.**
```
$ HOST_B64=$(base64 -w0 host.js)
$ sed "s|__HOST_JS_BASE64__|${HOST_B64}|" installers/install.sh > /tmp/install.release.sh
$ bash -n /tmp/install.release.sh                        # PARSE OK
$ grep -c '__HOST_JS_BASE64__' /tmp/install.release.sh   # 0
$ ENCODED=$(grep -oP 'LINKSHIT_HOST_JS_BASE64="\K[^"]+' /tmp/install.release.sh)
$ echo "$ENCODED" | base64 -d | diff -q - host.js        # ROUND-TRIP OK
```
The four checks the spec asks for all pass — but they don't catch finding #1.

**4. Finding #1 — the substitution bug, deterministic minimal repro.**
```
$ cat > /tmp/test-template.sh <<'TPL'
#!/usr/bin/env bash
readonly LINKSHIT_HOST_JS_BASE64="__HOST_JS_BASE64__"
if [ "${LINKSHIT_HOST_JS_BASE64}" = "__HOST_JS_BASE64__" ]; then
  echo "TEMPLATE_FORM"
else
  echo "EMBEDDED_FORM"
fi
TPL
$ sed "s|__HOST_JS_BASE64__|FAKE_BASE64_BLOB|" /tmp/test-template.sh > /tmp/test-template.released.sh
$ cat /tmp/test-template.released.sh
#!/usr/bin/env bash
readonly LINKSHIT_HOST_JS_BASE64="FAKE_BASE64_BLOB"
if [ "${LINKSHIT_HOST_JS_BASE64}" = "FAKE_BASE64_BLOB" ]; then     # ← both substituted
  echo "TEMPLATE_FORM"
else
  echo "EMBEDDED_FORM"
fi
$ bash /tmp/test-template.released.sh
TEMPLATE_FORM       # ← always takes the network branch
```
Confirmed against the real generated installer too (with a stub `claude` in PATH and an isolated `HOME`):
```
$ rm -rf /tmp/qa/home && mkdir -p /tmp/qa/home && HOME=/tmp/qa/home \
    PATH=/tmp/qa-stubs:$PATH yes y | bash /tmp/install.release.sh 2>&1 \
    | grep -E 'fetching|installed|ERROR'
Installing host.js into /tmp/qa/home/.linkshit/ ...
  fetching https://github.com/Replikanti/linkshit/releases/latest/download/host.js
ERROR: failed to download host.js from ...
```
Even with the embedded blob present, the released installer attempts a network fetch on every run. As long as the release also ships `host.js` as a separate asset (which it does), end users won't see a hard failure — but the entire point of the embedding is defeated, the README/comments lie about being self-contained, and it breaks for users running the installer in offline / restricted-network environments.

**Suggested fix** (no code change applied — QA only): change the comparison sentinel to a token that is NOT the same string as the substitution target. E.g.
```bash
# template
readonly LINKSHIT_HOST_JS_BASE64="__HOST_JS_BASE64__"
# ...
if [ "${LINKSHIT_HOST_JS_BASE64}" = "__HOST_JS_BASE64""__" ]; then    # split string, not a literal match
```
or simply test by length / pattern (`[ "${#LINKSHIT_HOST_JS_BASE64}" -lt 100 ]` since a real base64 of `host.js` is ~7600 chars).

**5. Finding #2 — `read` + `set -e` swallows EOF silently.**
```
$ HOME=/tmp/qa/home PATH=/tmp/qa-stubs:$PATH bash installers/install.sh </dev/null
==============================================================
  Linkshit — non-technical installer
==============================================================
✓ Node.js v22.22.0
✓ Claude Code: claude stub 0.0.0

Linkshit needs your Claude Code account to be already logged in
... (full prompt) ...
$ echo $?
1
```
The script exits non-zero with no error message. Recommend `read ... || logged_in=""` (or wrap in `if read ...; then ...; else ...; fi`) so EOF is treated as "no" with a friendly message.

**6. OS detection — correct.** `Linux*` and `Darwin*` paths set the expected manifest dir; anything else exits with a clear error to stderr.

**7. Manifest content — correct.** Simulated under isolated `HOME`:
```json
{
  "name": "com.replikanti.linkshit",
  "description": "Linkshit native messaging host",
  "path": "/tmp/qa/home/.linkshit/host.js",
  "type": "stdio",
  "allowed_origins": [
    "chrome-extension://pgcnimcldmdfkemofhjfnemieckciche/"
  ]
}
```
`python3 -m json.tool` parses cleanly. Trailing slash on `allowed_origins` matches Chrome's docs. Extension ID matches the pinned ID in `manifest.json`'s `key` field.

**8. Node version check — correct.** `node -v | sed 's/^v//' | cut -d. -f1` then `[ ... -lt 22 ]`. Tested with `v22.22.0` — accepted.

**9. Claude install gating — idempotent.** `if ! command -v claude` short-circuits when present.

**10. SVG mockups — well-formed and resolve.** `xml.etree` parses all three. GitHub markdown rendering (via `/markdown` API) emits `` for all three references.

**11. Extension still builds — yes, 3 files in zip.**
```
$ npm run pack
Your web extension is ready: .../web-ext-artifacts/linkshit-0.2.0.zip
$ unzip -l web-ext-artifacts/linkshit-0.2.0.zip
manifest.json (1069), background.js (1788), content.js (17419)
```

**12. PR scope — exactly the 7 declared files.**
```
$ git diff main..HEAD --name-only
.github/workflows/extras.yml
.github/workflows/release.yml
README.md
docs/screenshots/01-extensions-dev-mode.svg
docs/screenshots/02-load-unpacked.svg
docs/screenshots/03-panel-on-linkedin.svg
installers/install.sh
```
`host.js` and `manifest.json` untouched.

**13. No stray `score-server.js`, no smoke job in `extras.yml`.** Both confirmed via grep.

**14. release.yml — triggers `push` on `v*` tags only, `permissions: contents: write` minimal, `actions/checkout@v6` and `actions/setup-node@v6` current, `gh release create` with three asset args, no `--prerelease` flag. The `installers/install.release.sh#install.sh` rename trick is correctly used.**

**15. extras.yml — adds `shellcheck` job with `actions/checkout@v6` and runs `shellcheck installers/install.sh`. Clean.**

### What works

- `shellcheck installers/install.sh` clean, `bash -n` clean
- All three SVG mockups parse as valid XML; GitHub renders them as inline images via standard markdown image syntax
- README quick-start command (`bash ~/Downloads/install.sh`) matches what the installer expects
- Native messaging manifest content is byte-correct: name, path under `~/.linkshit/`, `type: stdio`, `allowed_origins` with trailing slash, pinned extension ID
- OS detection refuses unknown systems with a clear, multi-line stderr error
- Node version gate works (`v22.22.0` accepted)
- `claude` install step is gated on absence — re-runs are idempotent for that step
- Extension still builds: `npm run check / validate / lint / pack` all green; produced zip contains exactly the 3 expected files
- PR scope is exactly the 7 files declared; `host.js` and `manifest.json` untouched
- No `score-server.js` references reintroduced; no smoke job in `extras.yml` (correctly deferred to PR-C)
- All 8 CI checks green on the PR head
- All GitHub Actions pinned to current major (`@v6`); `lycheeverse/lychee-action` pinned to commit SHA
- Installer downloads only over HTTPS; no `curl … | bash`; no wildcard origin in the manifest

ylohnitram:
## QA Re-verification

### Verdict
✅ Pass

### Summary
All 3 actionable findings from the original QA report are genuinely closed by `882f86f`. End-to-end repro confirms the released installer now takes the EMBEDDED branch (no network fetch), `bash install.sh </dev/null` exits 0 with the friendly message, and the workflow's new length-check assertion fires loudly on a too-short blob. The defense-in-depth design also holds: even if the workflow's `sed` were silently regressed back to unanchored, the runtime length check in `install.sh` would still route correctly. CI is green on all 8 checks. No new issues introduced.

### Verification of previous findings

| # | Original concern | Fix verified | Result |
|---|------------------|--------------|--------|
| 1 | `sed` substitution clobbered the comparison literal, forcing every released installer to download `host.js` from the network on every run | Both layers exercised: (a) `install.sh` now detects template-vs-released by `${#LINKSHIT_HOST_JS_BASE64} -lt 200`, no string equality with the placeholder; (b) `release.yml` `sed` is anchored to `/^readonly LINKSHIT_HOST_JS_BASE64=/`. End-to-end run with stub `claude`/`node` and isolated `HOME` shows zero `fetching` log lines, installed `host.js` is byte-for-byte identical to source. Tested both layers independently — even a deliberately-buggy unanchored `sed` still produces a correct install thanks to the length check | ✅ |
| 2 | `set -euo pipefail` + `read` exited the script with code 1 and zero output if stdin was closed | `read … \|\| logged_in=""` now tolerates EOF; `bash install.sh </dev/null` prints the multi-line "log in first" message and exits 0. Verified | ✅ |
| 3 | Workflow's spot-check would not have caught Finding #1 because both checks were satisfied while the conditional was silently broken | `release.yml` "Embed host.js into install.sh" step now asserts `[ "${#ENCODED}" -lt 200 ]` and fails the build with an explanatory message. Simulated by feeding a 68-char short blob — the length-check fires, as expected | ✅ |
| 4 | (INFO) Unbounded glob in `mv "$(ls web-ext-artifacts/linkshit-*.zip)"` | Author marked as non-actionable nit; no change made. Confirmed unchanged in `release.yml` | ✅ |

### New findings

None. No regressions. No new issues introduced by the fixes.

### Detailed verification

**1. Files touched by `882f86f`:** exactly `installers/install.sh` and `.github/workflows/release.yml` — matches the claim in the commit message.

**2. shellcheck + parse:**
```sh
$ shellcheck installers/install.sh && bash -n installers/install.sh
# both clean
$ python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"
# parses
```

**3. End-to-end embed test:**
```sh
HOST_B64=$(base64 -w0 host.js)            # 7672 chars
sed "/^readonly LINKSHIT_HOST_JS_BASE64=/s|__HOST_JS_BASE64__|${HOST_B64}|" \
  installers/install.sh > install.release.sh
bash -n install.release.sh                # parses
grep -n '__HOST_JS_BASE64__' install.release.sh
# 121:# __HOST_JS_BASE64__ — if we used a string-equality test, ...
# (only the comment retained — the assignment line was substituted, the comment was not, anchored sed works as intended)
ENCODED=$(grep -oP '^readonly LINKSHIT_HOST_JS_BASE64="\K[^"]+' install.release.sh)
echo "${#ENCODED}"                        # 7672
echo "$ENCODED" | base64 -d | diff -q - host.js  # round-trip OK
```

Stub-driven release run:
```sh
mkdir -p qa-stubs
cat > qa-stubs/claude <<'EOF'
#!/bin/bash
[ "$1" = "--version" ] && echo "claude stub 0.0.0" || true
EOF
chmod +x qa-stubs/claude
ln -sf "$(which node)" qa-stubs/node

HOME=qa-test-home PATH=qa-stubs:$PATH \
  bash install.release.sh <<< 'y' 2>&1 | tee release-run.log
# exit 0
grep -c 'fetching' release-run.log         # 0  -> no network fallback
diff -q qa-test-home/.linkshit/host.js host.js
# Files qa-test-home/.linkshit/host.js and host.js are identical
```

**4. EOF tolerance:**
```sh
HOME=qa-test-home2 PATH=qa-stubs:$PATH bash installers/install.sh </dev/null
# ... prints the friendly "log in first, then re-run this installer." line
# exit: 0
```

**5. Workflow length-check assertion fires loudly on short blob:**
```sh
SHORT=$(head -c 50 /dev/urandom | base64 -w0)        # 68 chars
sed "/^readonly LINKSHIT_HOST_JS_BASE64=/s|__HOST_JS_BASE64__|${SHORT}|" \
  installers/install.sh > install.short.sh
ENCODED=$(grep -oP '^readonly LINKSHIT_HOST_JS_BASE64="\K[^"]+' install.short.sh)
[ "${#ENCODED}" -lt 200 ] && echo "would fail the build"  # would fail the build
```

**6. Defense-in-depth proof — runtime length check protects even if sed is unanchored again:**
```sh
HOST_B64=$(base64 -w0 host.js)
sed "s|__HOST_JS_BASE64__|${HOST_B64}|g" installers/install.sh > install.bug.sh
# Both placeholder occurrences substituted, including the comment on line 121
HOME=qa-bug-home PATH=qa-stubs:$PATH \
  bash install.bug.sh <<< 'y' 2>&1 | grep -E 'fetching|installed' | head
# host.js installed
diff -q qa-bug-home/.linkshit/host.js host.js   # identical
```
The length-based detection (`-lt 200`) makes the script robust to a regressed sed pattern. Good belt-and-braces.

**7. `case` statement against various `logged_in` values** (whitespace, empty, "yes", "no") — only `[Yy]*` matches the affirmative; everything else falls through to the friendly exit. No new edge case from the EOF tolerance.

**8. Sed anchor robustness:** line 32 in `installers/install.sh` is exactly `^readonly LINKSHIT_HOST_JS_BASE64="__HOST_JS_BASE64__"$` — no leading whitespace or tabs. The anchor `/^readonly LINKSHIT_HOST_JS_BASE64=/` matches reliably. Base64 alphabet is `[A-Za-z0-9+/=]` so the substituted blob can never contain the placeholder string `__HOST_JS_BASE64__` — no double-substitution risk inside the embedded payload.

**9. CI status (commit `882f86f`):**
```
Analyze JavaScript (javascript-typescript)   pass
CodeQL                                       pass
Markdown link check                          pass
SonarCloud Code Analysis                     pass
ci                                           pass
npm audit (high+)                            pass
shellcheck installer                         pass
web-ext lint                                 pass
```
All 8 checks green.

### What works

- Released installer now takes the EMBEDDED branch reliably (zero network fetches).
- Installed `host.js` is byte-for-byte identical to source.
- `bash install.sh </dev/null` (and Ctrl-D) prints a friendly message and exits 0.
- Anchored `sed` correctly leaves the explanatory comment on line 121 intact.
- Workflow's length-check assertion would catch a regressed embed.
- Defense-in-depth: runtime length check still routes correctly even if `sed` is later regressed to unanchored form.
- shellcheck clean, `bash -n` clean, YAML parses, all 8 CI checks pass.

### Tiny nit (no action needed)

The comment on line 124 of `installers/install.sh` says *"the unsubstituted placeholder is 21"* — actual length is 18 (`__HOST_JS_BASE64__`). Cosmetic, doesn't affect logic. Not raising as a finding.

PR #14, opened 2026-05-01
fix(release): rename install.release.sh -> install.sh on the release page

ylohnitram:
## Bug

v0.2.0 shipped with the installer named `install.release.sh` instead of `install.sh`. The README's Quick Start instructs:

```bash
bash ~/Downloads/install.sh
```

…which doesn't match what users actually download. The `install.sh` URL on v0.2.0 returns 404; only `install.release.sh` exists.

```
$ curl -sIL https://github.com/Replikanti/linkshit/releases/download/v0.2.0/install.sh | head -1
HTTP/2 404
$ curl -sIL https://github.com/Replikanti/linkshit/releases/download/v0.2.0/install.release.sh | head -1
HTTP/2 302
```

## Root cause

`release.yml`'s `gh release create` was given:

```
"installers/install.release.sh#install.sh"
```

The `#install.sh` was meant to rename the asset, but `#label` in `gh release create` is **only a UI display label** — the underlying asset filename comes from the path.

## Fix

Stage all three assets in a `release-assets/` directory under their final user-facing names **before** running `gh release create`. The release upload now matches the README.

## What this leaves open

`v0.2.0` itself is already published with the wrong asset name. Three options for cleanup, listed in declining destructiveness:

- **Delete the `v0.2.0` release + tag, then re-tag v0.2.0** after this PR merges. Cleanest, but destructive of a published release.
- **Tag `v0.2.1`** with the fix. Leaves a slightly-broken `v0.2.0` in history.
- **Edit the v0.2.0 release in-place**: download `install.release.sh`, re-upload as `install.sh`, delete the misnamed asset. Less destructive than full delete.

That decision goes to the human after CI passes.

## Test plan

- [ ] CI green

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,6 +77,17 @@ jobs:
           fi
           echo "✓ install.release.sh built and verified (${#ENCODED} chars embedded)"
 
+      - name: Stage assets with their final user-facing names
+        run: |
+          # gh release create's `path#label` syntax sets a UI label only —
+          # the actual asset filename comes from the path. README points
+          # users at install.sh, so rename the built artifact to that.
+          mkdir -p release-assets
+          cp "web-ext-artifacts/linkshit-extension-${GITHUB_REF_NAME}.zip" release-assets/
+          cp host.js release-assets/host.js
+          cp installers/install.release.sh release-assets/install.sh
+          ls -la release-assets/
+
       - name: Create GitHub Release with assets
         env:
           GH_TOKEN: ${{ github.token }}
@@ -84,6 +95,6 @@ jobs:
           gh release create "${GITHUB_REF_NAME}" \
             --title "${GITHUB_REF_NAME}" \
             --generate-notes \
-            "web-ext-artifacts/linkshit-extension-${GITHUB_REF_NAME}.zip" \
-            "host.js" \
-            "installers/install.release.sh#install.sh"
+            "release-assets/linkshit-extension-${GITHUB_REF_NAME}.zip" \
+            "release-assets/host.js" \
+            "release-assets/install.sh"


ylohnitram:
## QA Report

### Verdict
✅ Pass

### Summary
The fix correctly stages all three release assets under their final user-facing names before invoking `gh release create`, addressing the `install.release.sh` vs `install.sh` mismatch from v0.2.0. Diff is scoped to `.github/workflows/release.yml` only, YAML parses, all 9 CI checks are green, and a local dry-run of the staging step produces exactly the expected filenames.

### Findings

| 1 | Info | release.yml | YAML loads cleanly via `yaml.safe_load` | ✅ |
| 2 | Info | release.yml | Step ordering preserved: checkout → setup-node → npm ci → pack → embed → **stage (new)** → create release | ✅ |
| 3 | Info | release.yml | New Stage step uses the same `${GITHUB_REF_NAME}` env var convention as the rest | ✅ |
| 4 | Info | release.yml | Final asset filenames will be `linkshit-extension-vX.Y.Z.zip`, `host.js`, `install.sh` (matches README Quick Start) | ✅ |
| 5 | Info | release.yml | Embed step still produces `installers/install.release.sh` with all sanity checks (parses, placeholder gone, round-trip, length ≥200); Stage step copies that verified artifact to `release-assets/install.sh` — verifications remain in force | ✅ |
| 6 | Info | scope | `git diff main..HEAD --stat` shows only `.github/workflows/release.yml` (14 added, 3 removed); no incidental changes to `host.js`, `install.sh`, etc. | ✅ |
| 7 | Info | dry-run | Local replication of the staging step produces exactly `linkshit-extension-v0.2.0-test.zip`, `host.js`, `install.sh` in `release-assets/` | ✅ |
| 8 | Info | gitignore | `release-assets/` is not in `.gitignore`; moot because the directory only exists during a workflow run, never committed | ✅ |
| 9 | Info | UI | Old `path#label` syntax also set a UI display label; new syntax does not. No custom label is needed for this fix, but reviewers wanting one will need a different approach. Informational only. | ℹ️ |
| 10 | Info | CI | All 9 checks green on `75f4602` (ci, web-ext lint, npm audit, Markdown link check, shellcheck, native host smoke test, CodeQL JS, CodeQL aggregate, SonarCloud) | ✅ |

### Detailed verification

1. **YAML correctness**

   ```sh
   $ python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml')); print('YAML OK')"
   YAML OK
   ```

2. **Diff scope**

   ```sh
   $ git diff main..HEAD --stat
    .github/workflows/release.yml | 17 ++++++++++++++---
    1 file changed, 14 insertions(+), 3 deletions(-)
   ```

   Only the workflow file changed.

3. **Filename mapping (static read of new Stage step)**

   ```yaml
   mkdir -p release-assets
   cp "web-ext-artifacts/linkshit-extension-${GITHUB_REF_NAME}.zip" release-assets/
   cp host.js release-assets/host.js
   cp installers/install.release.sh release-assets/install.sh
   ```

   Then `gh release create` is given:
   - `release-assets/linkshit-extension-${GITHUB_REF_NAME}.zip`
   - `release-assets/host.js`
   - `release-assets/install.sh`

   Asset filenames on the Release will be exactly the basenames of these paths — matching the README's Quick Start (`Download install.sh from the latest release`).

4. **Embed step preserved**

   The Embed step (still present, unchanged) builds `installers/install.release.sh` and runs four guards: `bash -n`, placeholder-substitution check, base64 round-trip vs `host.js`, and length ≥200 chars (so the runtime template-detection branch in `install.sh` does not fall through to the network path). Stage step runs after that, so all guards still gate the artifact that ships as `install.sh`.

5. **Local dry-run of the staging step**

   ```sh
   $ export GITHUB_REF_NAME=v0.2.0-test
   $ mkdir -p /tmp/release-assets-test
   $ npm run pack > /dev/null 2>&1
   $ cp web-ext-artifacts/linkshit-*.zip /tmp/release-assets-test/linkshit-extension-${GITHUB_REF_NAME}.zip
   $ cp host.js /tmp/release-assets-test/host.js
   $ cp installers/install.sh /tmp/release-assets-test/install.sh   # stand-in for install.release.sh
   $ ls /tmp/release-assets-test/
   host.js  install.sh  linkshit-extension-v0.2.0-test.zip
   ```

   Three files, exactly the expected names.

6. **CI status**

   ```sh
   $ gh pr checks 14 --repo Replikanti/linkshit
   Analyze JavaScript (javascript-typescript)   pass
   CodeQL                                       pass
   Markdown link check                          pass
   SonarCloud Code Analysis                     pass
   ci                                           pass
   native host smoke test                       pass
   npm audit (high+)                            pass
   shellcheck installer                         pass
   web-ext lint                                 pass
   ```

### What works

- YAML is syntactically valid.
- Step ordering is correct; Stage runs after Embed and before release creation.
- Three asset paths map to exactly the three filenames the README points users to.
- Embed step's verifications (parse, placeholder check, round-trip, length ≥200) still gate the file that ships as `install.sh`.
- Diff is tightly scoped to one workflow file (14 / 3); no incidental edits.
- Local replication of the Stage step produces the expected three filenames.
- All 9 CI checks pass on the PR head.
- Comment in the new step explains the root cause (`gh`'s `path#label` is UI-only) so future maintainers won't reintroduce the bug.

PR #37, opened 2026-05-01
feat: skip-older-than time filter

### LINKED ISSUE #25: "Skip posts older than X hours" pre-filter
## Problem

There is no time-based pre-filter. Users doing a quick morning scan want
to ignore posts older than e.g. 24 hours; today the only way is to stop
the scroll early or rely on LinkedIn's own ordering (which mixes in stale
content via "in the news" / "you may have missed").

## Proposal

Add a new setting **Max age (hours)** (0 = unlimited, default = 0).
\`extractPost()\` reads the post's timestamp from the LinkedIn DOM
(typically a \`<time>\` element with a relative label like "2h" or
"3d"; fall back to parsing the visible text). If the post's age exceeds
the threshold, skip it the same way Promoted is skipped.

## Acceptance

- New numeric input in ⚙ next to the other filters.
- 0 (default) preserves current behaviour exactly.
- > 0 causes posts older than the threshold to be excluded from the
  queue (and ideally counted via a \`tooOld\` counter).
- Robust to LinkedIn's relative-time formatting (handles \`Nh\`, \`Nd\`,
  \`Nw\`, \`Nmo\`).

ylohnitram:
## Summary

- New \`Max age (hours, 0 = unlimited)\` setting in ⚙. Default 0 preserves current behaviour exactly.
- New \`extractAgeHours(el)\` reads LinkedIn's relative time string from \`.update-components-actor__sub-description\` (with a short-span fallback inside the actor block).
- Parser \`/(\d+)\s*(mo|min|[smhdwy])\b/i\` covers \`s\`, \`m\`/\`min\`, \`h\`, \`d\`, \`w\`, \`mo\`, \`y\`. Alternation order puts \`mo\` before \`m\` so \"1mo\" doesn't parse as 1 minute.
- New \`old\` counter (id \`lks-c-tooOld\`) shows how many posts were skipped for age.
- Unparseable age → no filtering (defensive).

Closes #25.

## Test plan

- [ ] CI green
- [ ] Code-level QA: filter only runs when \`maxAgeHours > 0\`; \`tooOld\` path skips \`dbPut\` / queue / captured bump.
- [ ] Parser correctness: \`mo\` before \`m\` in alternation prevents \"1mo\" → 1 minute mis-parse.

### DIFF
--- a/content.js
+++ b/content.js
@@ -119,15 +119,31 @@
   // LinkedIn renders post age as relative text (2h, 3d, 1mo, 2y) in the
   // actor sub-description. Returns hours or null when unparseable; null
   // is treated as "do not filter" (defensive — would rather over-include).
-  // Alternation order in the regex matters: "mo" must precede "m"
-  // otherwise "1mo" parses as 1 minute.
-  const AGE_UNITS = { s: 1 / 3600, min: 1 / 60, m: 1 / 60, h: 1, d: 24, w: 168, mo: 720, y: 8760 };
+  // The unit table starts with the longer keys so "mo" / "min" beat "m".
+  // String-based dispatch (instead of one alternation regex) keeps the
+  // hot path linear and avoids ReDoS-leaning patterns.
+  const AGE_UNITS = [
+    ['mo', 720],
+    ['min', 1 / 60],
+    ['s', 1 / 3600],
+    ['m', 1 / 60],
+    ['h', 1],
+    ['d', 24],
+    ['w', 168],
+    ['y', 8760],
+  ];
   function parseAgeHours(text) {
     if (!text) return null;
-    const m = text.match(/(\d+)\s*(mo|min|[smhdwy])\b/i);
-    if (!m) return null;
-    const factor = AGE_UNITS[m[2].toLowerCase()];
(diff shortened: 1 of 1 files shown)


ylohnitram:
## QA report

| Check | Result |
|---|---|
| Local \`check\` / \`validate\` / \`lint\` | ✅ pass |
| All 9 CI checks (after Sonar fix) | ✅ pass |
| Initial Sonar S5852 (ReDoS heuristic) | ❌ → ✅ mitigated by rewriting to digit-extract + longest-prefix table |
| Filter only runs when \`maxAgeHours > 0\` (default 0 unchanged) | ✅ |
| \`tooOld\` path skips \`dbPut\` / queue / captured bump | ✅ |
| Unparseable age returns null and never blocks | ✅ |
| Longer keys (\`mo\`, \`min\`) tried before single chars (\`m\`) — no \"1mo → 1 min\" mis-parse | ✅ — table iteration is in defined order, \`startsWith\` fires on first match |
| Live exercise on \`linkedin.com/feed\` with real timestamps | ⚠️ deferred to maintainer |

**Verdict: ✅ ship it.**

PR #40, opened 2026-05-01
feat: persist queued posts across tab close

### LINKED ISSUE #29: Persist queued-but-not-scored posts across tab close
## Problem

Captured posts that are waiting for the next scoring batch live only in
memory. Closing the tab (or LinkedIn forcing a reload) drops them, and
they get re-captured & re-queued from scratch on the next visit. That is
wasted scroll time.

## Proposal

When a post is captured and pre-filter passes, write it to IndexedDB
with \`status: 'queued'\` immediately (today they are only stored after
scoring). On panel mount, drain queued posts from DB into the in-memory
queue before the scroll loop starts.

## Acceptance

- Captured-and-pre-filter-passed posts visible in IndexedDB with
  \`status: 'queued'\`.
- After a tab close mid-session, reopening the page picks the queue
  back up — counters reflect carry-over.
- Already-scored posts are unaffected (they keep \`status: 'scored'\`).

ylohnitram:
## Summary

- Pre-filter-passing posts persist to IndexedDB with \`status: 'queued'\` instead of \`'new'\`.
- New \`dbAllQueued()\` helper queries via the existing \`status\` index.
- Boot drains queued records into the in-memory queue + \`seen\` set **before** starting the MutationObserver — eliminates the race where the observer could re-process a URN about to be loaded from disk.
- \`tryCapture()\` short-circuits on \`ex?.status === 'queued'\` (defense-in-depth — the \`seen.has()\` check earlier should already cover this, but extra safety guards against unexpected boot ordering).
- Status message on boot shows the carry-over count when non-zero.

Closes #29.

## Test plan

- [ ] CI green
- [ ] Code-level QA: drain happens before \`startObserver()\`; \`seen.add\` populated before \`queue.push\`; existing scoring path overwrites \`'queued'\` with \`'scored'\` via the same dbPut.
- [ ] Live exercise: capture posts, close tab mid-batch, reopen — queued posts surface in counter (deferred to maintainer).

### DIFF
--- a/content.js
+++ b/content.js
@@ -103,6 +103,11 @@
     const idx = db.transaction(STORE, 'readonly').objectStore(STORE).index('score');
     return cursorAll(idx, IDBKeyRange.lowerBound(min));
   }
+  async function dbAllQueued() {
+    const db = await dbReady;
+    const idx = db.transaction(STORE, 'readonly').objectStore(STORE).index('status');
+    return cursorAll(idx, IDBKeyRange.only('queued'));
+  }
 
   // ---------- Extractor ----------
   // LinkedIn rewrites these CSS class names occasionally. If captured count
@@ -283,12 +288,19 @@
       }
       return;
     }
-    await dbPut(post);
+    if (ex?.status === 'queued') {
+      // Already drained from disk into the in-memory queue at boot; nothing to do.
+      return;
+    }
     ui.bump('captured', 1);
     if (preFilter(post)) {
+      post.status = 'queued';
+      await dbPut(post);
       queue.push(post);
       ui.bump('queued', 1);
       maybeFlush();
+    } else {
+      await dbPut(post);
     }
   }
   function startObserver() {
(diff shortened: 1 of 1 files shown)


ylohnitram:
## QA report

| Check | Result |
|---|---|
| Local \`check\` / \`validate\` / \`lint\` | ✅ pass |
| All 9 CI checks | ✅ pass |
| Boot drain runs before \`startObserver()\` (race-safe) | ✅ verified ordering in diff |
| Drain populates \`seen\` so observer's later \`document.querySelectorAll\` re-pass is a no-op for those URNs | ✅ |
| Pre-filter pass: \`status: 'queued'\` is written; pre-filter fail: \`status: 'new'\` (preserved) | ✅ |
| Scoring path still rewrites status to \`'scored'\` (no behavioural change there — \`scoreBatch\` already returns \`status: 'scored'\`) | ✅ |
| Error path \`queue.unshift\` still works — DB record stays as \`'queued'\` for retry on next boot | ✅ |
| Clear DB still resets in-memory queue (untouched code path) | ✅ |
| Live exercise of close-tab-mid-batch | ⚠️ deferred to maintainer |

**Verdict: ✅ ship it.**

PR #47, opened 2026-05-01
fix: hide profile selector until 2+ profiles exist

### LINKED ISSUE #46: Profile dropdown is confusing with a single profile
## Problem

After v0.3.0, the panel header shows \`Linkshit · [Default]\` even when only one criteria profile exists. From a first-time user's perspective:

- The dropdown looks like part of the title, not a control.
- Clicking it reveals a single item (\"Default\") which feels like a bug.
- There's no label saying what \"Default\" refers to.

Real UX friction reported by maintainer after manual testing of v0.3.0.

## Proposal

- **Hide the profile selector entirely while exactly one profile exists.** Header reduces to \`Linkshit\` until the user creates a second profile in ⚙.
- **Once two or more profiles exist**, surface as \`Linkshit · Profile: [select]\` — the label \`Profile:\` makes the control self-explanatory.

\`refreshProfileSelector\` already runs on every profile change, so it just needs to toggle the wrapper's \`display\` and update the label visibility in one place.

## Acceptance

- Fresh install / single-profile users see no profile UI in the header.
- Creating a second profile via \`⚙ → New profile\` immediately reveals the selector + \`Profile:\` label.
- Deleting profiles back down to 1 hides the selector again.
- \`⚙ → New / Rename / Duplicate / Delete\` buttons still work the same way.

ylohnitram:
## Summary

- Wraps the header profile \`<select>\` in a span (\`#lks-profile-wrap\`) that defaults to \`display:none\` and includes a new \`Profile:\` label.
- \`refreshProfileSelector()\` toggles the wrap based on \`profiles.list.length > 1\`. Single source of truth — already called on load + every profile-mutating action.
- Header reads \`Linkshit\` for fresh users; flips to \`Linkshit · Profile: [select]\` once a second profile is created in ⚙.

Closes #46.

## Test plan

- [x] \`npm run smoke\` — 5/5
- [ ] CI green
- [ ] Code-level QA: wrap visibility computed in one place, no other code paths can leave the dropdown visible with a single profile.

### DIFF
--- a/content.js
+++ b/content.js
@@ -546,7 +546,7 @@
     panel.innerHTML = `
       <div id="lks-resize"></div>
       <header>
-        <span>Linkshit · <select id="lks-profile" title="Active profile"></select></span>
+        <span>Linkshit<span id="lks-profile-wrap" style="display:none"> · Profile: <select id="lks-profile" title="Active profile"></select></span></span>
         <span>
           <button id="lks-start" class="primary">Start</button>
           <button id="lks-pause">Pause</button>
@@ -682,6 +682,11 @@
         if (p.name === profiles.activeName) opt.selected = true;
         sel.append(opt);
       }
+      // Hide the whole "· Profile: [select]" affordance when a single profile
+      // exists — at that point the dropdown is just a confusing title-suffix
+      // with one inert option. It pops back the moment the user creates a
+      // second profile in ⚙.
+      $('lks-profile-wrap').style.display = profiles.list.length > 1 ? '' : 'none';
     }
     refreshProfileSelector();
 


ylohnitram:
## QA report

| Check | Result |
|---|---|
| Local \`check\` / \`validate\` / \`lint\` / \`smoke\` | ✅ pass |
| All 9 CI checks | ✅ pass |
| Wrap defaults to \`display:none\` so first paint is clean even before any JS runs | ✅ |
| Visibility computed in single source of truth (\`refreshProfileSelector\`) | ✅ |
| Initial \`refreshProfileSelector()\` call still in IIFE init | ✅ — wrap shows immediately if migration produced 2+ profiles |

**Verdict: ✅ ship it.**

PR #50, opened 2026-05-01
fix(manifest): drop background.scripts (MV2-only)

### LINKED ISSUE #49: Manifest warning: 'background.scripts' is MV2-only
## Symptom

Loading the unpacked extension in Chrome MV3 surfaces a warning:

> 'background.scripts' requires manifest version of 2 or lower.

The extension still loads, but the warning is real — \`manifest.json\` carries both \`service_worker\` (MV3) and \`scripts\` (MV2) under the \`background\` block. Chrome ignores \`scripts\` in MV3 but reports the inconsistency.

## Fix

Drop \`background.scripts\`. Keep \`service_worker\` + \`type: \"module\"\` (the latter is valid for MV3 service workers — declares the worker as an ES module).

## Test surface

\`npm run validate\` (manifest JSON shape), \`web-ext lint\`, manual reload in Chrome.

ylohnitram:
Removes the legacy \`background.scripts\` key. \`service_worker\` + \`type: \"module\"\` stay — both valid in MV3.

Closes #49.

## Verification

- [x] \`npm run check\` / \`validate\` / \`lint\` / \`pack\` — clean
- [ ] CI green
- [ ] Code-level QA: only the one key dropped; service worker still pointed at \`background.js\`.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -7,7 +7,6 @@
   "permissions": ["nativeMessaging"],
   "background": {
     "service_worker": "background.js",
-    "scripts": ["background.js"],
     "type": "module"
   },
   "content_scripts": [


ylohnitram:
Reverting — `background.scripts` is a deliberate cross-browser shim. Firefox MV3 requires it as a fallback for browsers that don't yet support `service_worker`; Chrome ignores it but emits the harmless info-warning we saw. `web-ext lint` enforces this with rule BACKGROUND_SERVICE_WORKER_NOFALLBACK.

PR #55, opened 2026-05-02
fix: extract real author and observe whole feed subtree

### LINKED ISSUE #54: extractPost picks avatar link as author, returns null for every post (captured stays 0 after PR #52)
## Symptom

After installing v0.3.2 and opening `linkedin.com/feed/`, the side panel sits at:

```
captured 0
promoted 0
old 0
queued 0
scored 0
hits 0
End of feed.
```

despite real posts being present in the DOM.

## Diagnosis

Selectors from PR #52 still match the live DOM (verified via DevTools):

```js
({
  feed: document.querySelectorAll('[data-testid="mainFeed"]').length,           // 1
  posts: document.querySelectorAll('[data-testid="expandable-text-box"]').length, // 71
  authors: document.querySelectorAll('a[href*="/in/"], a[href*="/company/"]').length, // 160
})
```

So `[data-testid="mainFeed"]` and `[data-testid="expandable-text-box"]` still anchor correctly. The bug is downstream.

### Bug 1 — author selector matches the avatar wrapper first

Each post has ~2 author links (160 / 71 ≈ 2.25). The first one in DOM order wraps the avatar `` and has empty `innerText`:

```js
// On a real wrapper:
{ authorHref: '...linkedin.com/in/ylohnitram/', authorInnerText: '""', authorContainsImg: true }
```

`extractPost()` does:

```js
const authorRaw = (authorLink.innerText || '').split('\n').map(l => l.trim()).filter(Boolean);
const author = authorRaw[0];
if (!author) return null;
```

`authorRaw` is `[]` → `author` is `undefined` → returns `null` for every post. `captured` never bumps.

### Bug 2 — observer architecture assumes posts are direct children of feed

PR #52 stated *"observes only its direct children — replaces the previous subtree-wide URN-attribute selector match"*. But the live DOM has 12 direct children of `[data-testid="mainFeed"]` and 71 posts nested **8 levels deep**:

```js
({ directChildren: 12, postsInsideFeed: 71, firstPostDepthFromFeed: 8 })
```

Even after Bug 1 is fixed, `childList: true` on the feed will only fire when one of those 12 wrappers gets added/removed, not when LinkedIn lazy-mounts more posts inside an existing wrapper. New posts loaded by scroll would be missed.

## Why PR #52 didn't catch this

The PR description's test plan had `Live exercise on real linkedin.com/feed/ (deferred to maintainer)` — the fix passed `npm run smoke` (which doesn't touch the DOM) but was never exercised against real LinkedIn. Both bugs require a live page to surface.

## Fix surface

`content.js` only:

- Author extraction: pick the first author link inside the wrapper whose `innerText` is non-empty (skip avatar wrappers).
- Capture pipeline: iterate `feed.querySelectorAll('[data-testid="expandable-text-box"]')` and walk up to the smallest ancestor that also contains an author link. That's the post wrapper. Pass it to `extractPost`.
- Observer: keep watching `feed` but with `subtree: true`, debounce a rescan over text-boxes inside `feed`. Dedup via the existing djb2 urn key handles re-processing safely.

## Test plan

- `npm run check` / `validate` / `lint` / `smoke` clean
- Live exercise on `linkedin.com/feed/`: `captured` should jump within seconds of mount; queued/scored should follow with default criteria

ylohnitram:
Two regressions from PR #52 that only surfaced once a maintainer ran v0.3.2 against real `linkedin.com/feed/`. Closes #54.

## Bug 1 — `extractPost` returned null for every post

`AUTHOR_SEL = 'a[href*="/in/"], a[href*="/company/"]'` matches both the avatar `<a>` (wraps ``, empty `innerText`) and the name link. `el.querySelector(AUTHOR_SEL)` returns the first match in DOM order — the avatar — so `authorRaw[0]` was always `undefined` and the function exited at `if (!author) return null`. `captured` stayed pinned at 0.

Fix: iterate `el.querySelectorAll(AUTHOR_SEL)` and take the first link whose `innerText` has a non-empty first line.

## Bug 2 — observer missed every lazy-mounted post

PR #52 said *"observes only its direct children"* and the seed loop walked `feed.children`. The live DOM has **12 direct children** with **71 posts nested ~8 levels deep**. Even after Bug 1 is fixed, `{ childList: true }` on `feed` only fires when one of those 12 wrappers is added/removed; lazy-mounts inside an existing wrapper are silent.

Fix: switch to `{ childList: true, subtree: true }` and on each debounced (150 ms) batch rescan `feed.querySelectorAll(POST_TEXT_SEL)`, walking up from each text-box via the new `findWrapper(textBox, feed)` helper. The existing djb2 urn dedup makes rescans cheap and idempotent.

## Diagnostic data

```js
({
  feed: 1,
  postsInsideFeed: 71,
  authors: 160,                  // ~2.25 per post — avatar + name
  directChildren: 12,
  firstPostDepthFromFeed: 8,
})
// On a real wrapper, first AUTHOR_SEL match:
({ authorContainsImg: true, authorInnerText: '""' })
```

## Test plan

- [x] `npm run check` / `validate` / `lint` / `smoke` — clean, smoke 5/5
- [ ] CI green
- [ ] Live exercise on `linkedin.com/feed/`: `captured` jumps within seconds of mount; with default criteria, `queued` and `scored` follow
- [ ] Regression: sponsored posts still register as `promoted` (path unchanged, exercised via the new wrapper)

### DIFF
--- a/content.js
+++ b/content.js
@@ -221,14 +221,33 @@
     }
     return null;
   }
+  // Walk up from a post-text element to the smallest ancestor that also
+  // contains an author link — that is the post wrapper. Stops at feed
+  // itself so we never escape into the page chrome. Used by capture to
+  // turn each [data-testid="expandable-text-box"] into a wrapper that
+  // extractPost / isPromoted / extractAgeHours all operate on.
+  function findWrapper(textBox, feed) {
+    let n = textBox.parentElement;
+    while (n && n !== feed) {
+      if (n.querySelector(AUTHOR_SEL)) return n;
+      n = n.parentElement;
+    }
+    return null;
+  }
   function extractPost(el) {
     const textEl = el.querySelector(POST_TEXT_SEL);
     if (!textEl) return null;
-    const authorLink = el.querySelector(AUTHOR_SEL);
+    // Each post carries ~2 author links: an avatar wrapper around an
+    // <img> (first in DOM order, empty innerText) and the name link.
+    // Pick the first link whose innerText has a non-empty first line so
+    // the avatar wrapper does not poison the author field.
+    let authorLink = null, authorRaw = null;
+    for (const a of el.querySelectorAll(AUTHOR_SEL)) {
+      const raw = (a.innerText || '').split('\n').map(l => l.trim()).filter(Boolean);
+      if (raw[0]) { authorLink = a; authorRaw = raw; break; }
+    }
     if (!authorLink) return null;
-    const authorRaw = (authorLink.innerText || '').split('\n').map(l => l.trim()).filter(Boolean);
     const author = authorRaw[0];
-    if (!author) return null;
     const text = (textEl.innerText || '').trim();
     if (!text || text.length < 30) return null;
     const subtitle = authorRaw.slice(1).join(' · ').slice(0, 200);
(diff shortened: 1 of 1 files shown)


ylohnitram:
## QA report

### Static review

| # | Check | Result |
|---|-------|--------|
| 1 | `npm run check` / `validate` / `lint` / `smoke` | ✅ smoke 5/5 |
| 2 | All CI workflows on PR HEAD | ✅ green |
| 3 | `extractPost` author guard now skips empty-text links | ✅ loop bails on `raw[0]` truthy; falsy `innerText` no longer poisons the field |
| 4 | `findWrapper` stops at `feed`, never escapes | ✅ `while (n && n !== feed)` |
| 5 | `findWrapper` returns null for orphan text-boxes (e.g. text-box rendered without an author link in any ancestor) | ✅ returns null → `tryCapture` skipped, no throw |
| 6 | Observer rescan is debounced and idempotent | ✅ `pending` flag + 150 ms timeout; djb2 urn dedup means re-running `tryCapture(wrapper)` on the same wrapper is a no-op after the first call |
| 7 | `isPromoted` / `extractAgeHours` paths still operate on the wrapper, not the text-box | ✅ unchanged — wrapper passed in is the same shape (an ancestor that contains both author and text) |
| 8 | No new runtime deps | ✅ no `package.json` changes |
| 9 | Comment block at `extractor` section reflects the new shape | ✅ `startObserver` comment rewritten; the higher-level \"selector drift\" guidance at L147–154 is still accurate |

### Live exercise

Run on a real `linkedin.com/feed/` session via Option B (extractor logic walked manually in DevTools):

```js
({ totalTextBoxes: 71, captured: 71, noWrapper: 0, noAuthor: 0 })
```

Every text-box resolves to a wrapper, every wrapper has a non-empty author. ✅ The bug that pinned `captured` to 0 in v0.3.2 is gone.

### Verdict

✅ Ready to merge. Static, CI, and live all green.

PR #61, opened 2026-05-02
fix: scroll <main> instead of document.body for infinite-feed loop

### LINKED ISSUE #60: Auto-scroller targets document.body but real scroll container is <main>; 'End of feed' fires after initial render on 2026-05 DOM
v0.3.4 captures the initial 21 or so posts that LinkedIn renders on first paint, then immediately stops with \"End of feed.\" — the auto-scroller's height-stable check trips because \`document.body.scrollHeight\` never moves. Live diagnostic confirms \`document.body\` is not the scrollable element on the new DOM.

## Diagnosis

\`\`\`js
({ before: document.body.scrollHeight,
   after:  (window.scrollTo(0, document.body.scrollHeight), document.body.scrollHeight) })
// → { before: 927, after: 927 }
\`\`\`

927 px is roughly viewport height; the body has effectively no scrollable content. The actual scrollable element is \`<main>\`:

\`\`\`
{ tag: 'MAIN', scrollH: 24736, clientH: 875 }   // ~28x viewport
\`\`\`

The 2026-05 React rewrite moved scroll to a flex-grow \`<main>\` inside a fixed-height shell. \`window.scrollTo(0, document.body.scrollHeight)\` is a no-op.

## Bug

\`content.js\` scroller at the \`tick\` function:

\`\`\`js
window.scrollTo(0, document.body.scrollHeight);
...
if (document.body.scrollHeight === lastH) {
  if (++stable > 6) { stopScroll(); ui.setStatus('End of feed.'); ... }
}
\`\`\`

Both the scroll target and the height watermark use \`document.body\`. On the new DOM \`body.scrollHeight\` is constant at ~viewport, so \`stable\` reaches 6 within a few ticks and the loop exits with \"End of feed.\" — silently false.

## Fix

Resolve the actual scroll container at tick time and use it for both the scroll call and the height watermark:

\`\`\`js
const scroller = document.querySelector('main') || document.scrollingElement || document.body;
scroller.scrollTo(0, scroller.scrollHeight);
...
if (scroller.scrollHeight === lastH) { ... }
\`\`\`

Re-resolving each tick handles React unmount/remount of \`<main>\`. The fallback chain keeps the old behaviour on any future LinkedIn variant that goes back to body-scroll.

## Test plan

- \`npm run check\` / \`validate\` / \`lint\` / \`smoke\` — no behaviour change to host plumbing, must stay green.
- Live exercise on real \`linkedin.com/feed/\`: clicking Start should scroll past the initial render and \`captured\` should keep climbing for several minutes (not stop after a few ticks). \`End of feed.\` should now fire only when \`<main>\` actually stops growing.

## Why earlier verification missed this

PR #58's live test caught the capture-pipeline regression but the maintainer never clicked Start on that run — they only verified that initial-render posts were captured. The scroll loop was untouched since v0.3.0 and assumed body-scroll, which was correct on the old DOM but silently broke with PR #52's React rewrite.

ylohnitram:
The 2026-05 React rewrite moved scroll out of \`document.body\` into a flex-grow \`<main>\`. The auto-scroller's \`window.scrollTo(0, body.scrollHeight)\` is a no-op on that DOM and the height-stable end-detection trips on the constant \`body.scrollHeight\` — \`captured\` plateaus at the initial render (~21 posts) and the panel reports \"End of feed.\" within seconds of clicking Start. Closes #60.

## Diagnosis

Run on real \`linkedin.com/feed/\`:

\`\`\`js
({ before: document.body.scrollHeight,
   after:  (window.scrollTo(0, document.body.scrollHeight), document.body.scrollHeight) })
// → { before: 927, after: 927 }
\`\`\`

927 px ≈ viewport. The actual scrollable element:

\`\`\`
{ tag: 'MAIN', scrollH: 24736, clientH: 875 }   // ~28x viewport, the real feed
\`\`\`

## Fix

Resolve the scroll container per tick:

\`\`\`js
const scroller = document.querySelector('main') || document.scrollingElement || document.body;
scroller.scrollTo(0, scroller.scrollHeight);
// ...
if (scroller.scrollHeight === lastH) { ... } else { lastH = scroller.scrollHeight; ... }
\`\`\`

Re-resolving each tick is one selector match and survives React re-mount of \`<main>\`. The fallback chain keeps the old behaviour on any LinkedIn variant that scrolls the document.

\`lastH\` initialization at \`let lastH = 0\` stays — first tick reads the real height, sets stable=0, loop continues.

## Test plan

- [ ] CI green
- [ ] Live exercise on \`linkedin.com/feed/\`: click Start, expect \`captured\` to keep climbing well past the initial render. \"End of feed.\" should fire only after \`<main>\` truly stops growing (typically minutes).
- Regression: pause / resume, maxPosts cap, panel state machine — paths unchanged.

### DIFF
--- a/content.js
+++ b/content.js
@@ -490,16 +490,24 @@
     }
     const tick = () => {
       if (!scrollTimer) return;
-      window.scrollTo(0, document.body.scrollHeight);
+      // The 2026-05 React rewrite moved scroll out of document.body into a
+      // flex-grow <main>; window.scrollTo / body.scrollHeight are no-ops on
+      // that DOM and the height-stable check trips after the initial render.
+      // Resolve the real container per tick (cheap, survives <main>
+      // re-mount) and fall back to the document scroller / body so older
+      // or future LinkedIn variants that scroll the document still work.
+      const scroller = document.querySelector('main') || document.scrollingElement || document.body;
+      scroller.scrollTo(0, scroller.scrollHeight);
       document.querySelectorAll('button').forEach(b => {
         const t = (b.innerText || '').toLowerCase();
         if (t.includes('load more comment') || t.includes('show more replies')
          || t.includes('zobrazit další') || t.includes('načíst předchozí')) b.click();
       });
-      if (document.body.scrollHeight === lastH) {
+      const h = scroller.scrollHeight;
+      if (h === lastH) {
         if (++stable > 6) { stopScroll(); ui.setStatus('End of feed.'); maybeFlush(true); return; }
       } else {
-        stable = 0; lastH = document.body.scrollHeight;
+        stable = 0; lastH = h;
       }
       if (seen.size - postsAtStart >= CFG.maxPosts) {
         stopScroll(); ui.setStatus(`maxPosts=${CFG.maxPosts} reached.`); maybeFlush(true); return;


ylohnitram:
## QA report

### Static review

| # | Check | Result |
|---|-------|--------|
| 3 | Scroll target uses scroller, height watermark uses scroller (no \`document.body\` left in tick) | ✅ \`grep document.body\` in \`tick\` returns 0 |
| 4 | Scroller fallback chain matches LinkedIn variants | ✅ \`<main>\` (current) → \`document.scrollingElement\` (historical document scroll) → \`document.body\` (last resort) |
| 5 | \`lastH\` initial 0 still works on first tick | ✅ first tick reads real height \>\> 0, sets stable=0 |
| 6 | maxPosts cap path unchanged | ✅ \`seen.size - postsAtStart >= CFG.maxPosts\` untouched |
| 7 | Pause / Resume preserves the new scroll target | ✅ scroller is resolved each tick, no captured reference to old container |
| 8 | Auto-clicked button list (\"load more comment\" / \"zobrazit další\" / etc.) untouched | ✅ regression-safe |

### Live exercise

Required before merge: click Start on \`linkedin.com/feed/\` and observe \`captured\` keep climbing past 21. Maintainer is already running this in interactive testing.

### Verdict

✅ Static and CI green. Awaiting live confirmation from maintainer.

PR #65, opened 2026-05-02
fix: scored counter accounts for previously-scored re-encounters

### LINKED ISSUE #63: Tab crash around captured ~230 + scored counter does not advance while hits is non-zero

## Suggested investigation

## Out of scope here

ylohnitram:
The \`scored\` counter only advanced after a fresh LLM batch returned. Two paths that surface previously-stored hits — the boot history-render loop and \`tryCapture\`'s \`ex?.status === 'scored'\` branch — render results via \`ui.addResult\` (which bumps \`hits\`) without ever bumping \`scored\`.

Maintainer reading: \`captured 228, queued 77, scored 0, hits 3\` — three boot-rendered hits, zero LLM batches yet, counter stuck at 0. Violates the obvious invariant \`hits ≤ scored\`.

Closes #63 — Symptom B only. Symptom A (tab crash near captured ~230) stays open and needs a Performance-tab capture before any fix.

## Fix

\`content.js\` only.

\`\`\`diff
-    const past = await dbAllScored(lowerBound);
-    for (const p of past.slice(0, 50)) {
-      ui.addResult(p.score >= CFG.scoreThresh ? p : { ...p, borderline: true });
-    }
+    const past = await dbAllScored(lowerBound);
+    for (const p of past.slice(0, 50)) {
+      ui.bump('scored', 1);
+      ui.addResult(p.score >= CFG.scoreThresh ? p : { ...p, borderline: true });
+    }
\`\`\`

\`\`\`diff
       const ex = await dbGet(post.urn);
       if (ex?.status === 'scored') {
+        ui.bump('scored', 1);
         if (ex.score >= CFG.scoreThresh) ui.addResult(ex);
         else if (CFG.showBorderline && ex.score >= CFG.scoreThresh - CFG.borderlineDelta) {
           ui.addResult({ ...ex, borderline: true });
         }
         return;
       }
\`\`\`

Re-encounter bump fires regardless of whether the re-encountered score crosses the threshold — it still counts as a post we have a score for. \`maybeFlush\` and rescore continue to bump \`scored\` by \`scored.length\` per LLM batch; those paths are unchanged.

## Test plan

- [ ] CI green
- [ ] Live: open the tab on \`linkedin.com/feed/\` with stored hits in IDB. Boot should bump \`scored\` to match \`hits\` immediately. After scrolling, \`scored\` should advance for both fresh batches and re-encountered previously-scored posts.

Out of scope: any change to the math definition of \`captured\` / \`queued\` / \`promoted\`. Tooltip / labeling work for the counter row is bundled into the broader UX ticket.

### DIFF
--- a/content.js
+++ b/content.js
@@ -422,6 +422,10 @@
       }
       const ex = await dbGet(post.urn);
       if (ex?.status === 'scored') {
+        // Re-encountering a post the LLM already scored in a previous
+        // session counts toward `scored` so the panel invariant
+        // hits ≤ scored holds across reloads.
+        ui.bump('scored', 1);
         if (ex.score >= CFG.scoreThresh) ui.addResult(ex);
         else if (CFG.showBorderline && ex.score >= CFG.scoreThresh - CFG.borderlineDelta) {
           ui.addResult({ ...ex, borderline: true });
@@ -998,6 +1002,9 @@
       : CFG.scoreThresh;
     const past = await dbAllScored(lowerBound);
     for (const p of past.slice(0, 50)) {
+      // Bump `scored` for each past hit we surface so the panel invariant
+      // hits ≤ scored holds at boot before any fresh scoring runs.
+      ui.bump('scored', 1);
       ui.addResult(p.score >= CFG.scoreThresh ? p : { ...p, borderline: true });
     }
     const anyStored = await dbAllScored(0);


ylohnitram:
## QA report

### Static review

| # | Check | Result |
|---|-------|--------|
| 3 | Bump in \`tryCapture\` scored branch fires once per re-encounter regardless of threshold | ✅ unconditional inside the branch |
| 4 | Bump in boot loop fires once per rendered past post | ✅ inside the slice(0, 50) iteration |
| 5 | \`maybeFlush\` still bumps \`scored\` by \`scored.length\` | ✅ untouched |
| 6 | \`addResult\` still bumps \`hits\` (only if !borderline) | ✅ untouched |
| 7 | Invariant hits ≤ scored holds for: boot only, observer re-encounter, fresh LLM batch | ✅ each path bumps scored at least as many times as addResult bumps hits |
| 8 | No double-counting in maybeFlush (scored bumped once per batch via batch.length, not per addResult) | ✅ paths are disjoint |

### Live exercise — pending maintainer

Open the tab on \`linkedin.com/feed/\` with stored hits already in IDB:

- Before scrolling: \`scored\` should equal the boot-rendered hits + borderlines (not 0).
- During scrolling: re-encountering previously-scored posts (dedup hits in IDB \`scored\` status) advances \`scored\`. Fresh batches advance \`scored\` and \`hits\` together.

Verdict: ✅ ready to merge once live confirms.

PR #70, opened 2026-05-02
fix: cap visible result cards at 200 to bound panel DOM growth

ylohnitram:
Third mitigation candidate for #63 Symptom A. PR #66 narrowed observer scope; PR #68 added marker-skip in rescan to bound async fan-out. Both helped (v0.3.7 lasted significantly longer before crashing) but the crash still recurs at extended sessions.

Remaining unbounded sink: the result card panel. Every \`addResult\` call appends a card to \`#lks-body\` — no cap, no virtualization. Long sessions accumulate hundreds of cards (boot history loop renders up to 50 immediately, then every fresh hit, every previously-scored re-encounter, every borderline on top). Each card is innerHTML + 4 child elements + 240-char snippet text. The DOM cost compounds with whatever else accumulates over time and eventually pushes the renderer over the edge.

## Fix

Cap at \`MAX_PANEL_RESULTS = 200\`. Insert is score-descending; when count exceeds the cap, evict the last (lowest-score) card.

\`\`\`js
const MAX_PANEL_RESULTS = 200;
function addResult(post) {
  // ... existing insert logic with div.dataset.urn = post.urn ...
  while (body.children.length > MAX_PANEL_RESULTS) {
    const last = body.lastElementChild;
    if (last.dataset.urn) rendered.delete(last.dataset.urn);
    last.remove();
  }
}
\`\`\`

The evicted post still exists in IDB (\`dbAllScored\` returns it), so the history view shows everything. The \`seen\` Set in \`tryCapture\` prevents the urn from being re-walked this session, so the eviction doesn't cause card thrashing.

Counters \`hits\` and \`scored\` do NOT decrement on eviction — they reflect \"matches found this session,\" not \"currently displayed.\"

## Test plan

- [ ] CI green
- [ ] Live: maintainer runs an extended session — past whatever threshold the previous build crashed at. Expect the panel to stabilize at exactly 200 cards visible after enough hits surface.

## If this still crashes

Then the crash isn't the panel either, and we need a Performance-tab capture before any further blind mitigation. Open candidates after this exhaust: in-memory queue back-pressure, Chrome MV3 SW lifecycle issues with connectNative, IDB transaction queue depth.

Mitigates #63 Symptom A.

### DIFF
--- a/content.js
+++ b/content.js
@@ -996,11 +996,19 @@
       if (el) el.textContent = counters[name];
       if (name === 'hits' || name === 'scored') hideHint();
     }
+    // Cap visible result cards to bound panel DOM growth in long sessions.
+    // History view (dbAllScored) still surfaces everything stored. Insert
+    // order is score-descending; eviction drops the lowest-score card from
+    // the tail. seen.add prevents tryCapture from re-walking the urn within
+    // this session, so evicted posts only re-render via the boot history
+    // loop on a future page load.
+    const MAX_PANEL_RESULTS = 200;
     function addResult(post) {
       if (rendered.has(post.urn)) return;
       rendered.add(post.urn);
       if (!post.borderline) bump('hits', 1);
       const div = document.createElement('div');
+      div.dataset.urn = post.urn;
       div.className = post.borderline ? 'result borderline' : 'result';
       div.innerHTML = `
         <div><span class="score"></span><span class="author"></span></div>
@@ -1017,6 +1025,11 @@
         c => Number.parseInt(c.querySelector('.score').textContent, 10) < post.score
       );
       if (after) after.before(div); else body.append(div);
+      while (body.children.length > MAX_PANEL_RESULTS) {
+        const last = body.lastElementChild;
+        if (last.dataset.urn) rendered.delete(last.dataset.urn);
+        last.remove();
+      }
     }
     return { setStatus, bump, addResult, showHint, hideHint, refreshControls, clearResults };
   })();


ylohnitram:
## QA report

### Static review

| # | Check | Result |
|---|-------|--------|
| 3 | \`div.dataset.urn\` set before insert so the evictor can read it | ✅ stamped on the new \`div\` ahead of placement |
| 4 | Eviction loop is \`while\`, not \`if\`, in case insert pushes count by more than one | ✅ \`while (body.children.length > MAX_PANEL_RESULTS)\` |
| 5 | \`rendered.delete(urn)\` on evict so the urn is no longer considered \"in panel\" | ✅ ahead of \`last.remove()\` |
| 6 | Eviction targets the last child (lowest score under desc sort) | ✅ \`body.lastElementChild\` |
| 7 | Counters \`hits\` / \`scored\` do NOT decrement on eviction | ✅ counter math untouched in eviction path |
| 8 | History view still shows evicted posts via \`dbAllScored\` | ✅ \`addResult\` only manages live panel; IDB store is unchanged |
| 9 | \`seen\` Set still prevents the same urn being re-processed by \`tryCapture\` after eviction | ✅ \`seen.add\` happens in \`tryCapture\`, separate from \`rendered\` |
| 10 | No regression on borderline rendering or score-descending sort order | ✅ insert path unchanged, eviction is post-insert |

### Live exercise

Maintainer's install dir will be patched. Static + CI green is sufficient to merge per maintainer authorization. Live confirmation lands as a follow-up on #63: if this is the last contributing factor, the crash should not recur over an extended session. If it does, we need a Performance-tab capture before any further blind change.

### Verdict

PR #72, opened 2026-05-02
fix: bound scoring retries and pause work on hidden tabs

### LINKED ISSUE #63: Tab crash around captured ~230 + scored counter does not advance while hits is non-zero

ylohnitram:
Two compounding sources of CPU spin observed by the maintainer in extended-session testing of v0.3.8: laptop fans audible, Chrome flagged the tab as \"using extra resources\" while in background, and the tab was unresponsive on refocus until contents re-rendered. Closes #63.

## Bug 1 — \`maybeFlush\` retries failed batches forever

\`content.js:340-356\` (pre-fix):

\`\`\`js
} catch (e) {
  for (const p of batch) queue.unshift(p);   // batch back in queue
  if (e.code === 'quota_exhausted') { /* pause flag */ }
  else { ui.setStatus(...); await sleep(15000); }
} finally {
  scoring = false;
  if (queue.length === 0) pendingForce = false;
  if (!quotaPaused && (queue.length >= CFG.batchSize || ...)) {
    maybeFlush();   // <-- retry the same batch
  }
}
\`\`\`

Non-quota errors put the batch back in the queue, slept 15 s, then the finally block recursed into \`maybeFlush()\` with the same batch waiting. In a background tab where Chrome MV3 kills the service worker after ~30 s of idleness, every flush fails with a port-disconnected error. Each cycle = 15 s pause + failed scoring attempt + recursion. Chrome throttles \`setTimeout\` to ~1/min in hidden tabs but the recursive promise chain doesn't care, leading to a thunderstorm of pending work that drops onto the main thread the moment the user refocuses.

**Fix:** count consecutive non-quota failures. After \`MAX_CONSECUTIVE_ERRORS = 3\`, set a \`scoringStalled\` flag and surface \"Scoring stalled after N errors; click Resume.\" Reset on a successful batch or manual Start/Resume. Mirrors the existing quota stall UX.

## Bug 2 — auto-scroll and \`maybeFlush\` ignored tab visibility

The tick scheduled itself every 3-6 s regardless of whether the user was looking at the tab. In hidden tabs Chrome throttles \`setTimeout\` to ~1/min, but each tick still does a full scroll + \`document.querySelectorAll('button').forEach\` + scrollHeight check, plus everything the resulting feed mount triggers via the MutationObserver and the queue/maybeFlush pipeline. None of the captured posts are useful while the user is elsewhere; the work compounds with Bug 1.

**Fix:** \`maybeFlush\` early-exits on \`document.visibilityState === 'hidden'\`. The auto-scroll tick early-exits but re-queues itself so the loop resumes the moment the tab comes back. A \`visibilitychange\` listener kicks one \`maybeFlush()\` on visible to drain any queued posts.

## Test plan

- [ ] CI green
- [ ] Live: extended session including \"switch tab away, come back\" cycles. Expect: tab not warming, no Chrome \"using extra resources\" warning, no unresponsive on refocus. If a real scoring error happens (host crash, quota), expect status to settle on \"Scoring stalled\" rather than spinning.

## If this doesn't help

Performance profile is required before any further blind work. Past this point the search space is too wide to keep guessing — too many candidates, all small.

### DIFF
--- a/content.js
+++ b/content.js
@@ -317,16 +317,32 @@
   // the auto re-flush at the end of maybeFlush would immediately retry and
   // burn the next request against the still-exhausted limit.
   let quotaPaused = false;
+  // Bounded retry on non-quota scoring errors. Without this the catch path
+  // re-queues the failed batch, sleeps 15 s, and the finally block calls
+  // maybeFlush() recursively — same batch fails again, every 15 s, forever.
+  // In a background tab where Chrome MV3 kills the service worker, every
+  // batch fails until the user refocuses, and the deferred retries fire in
+  // a thunderstorm when they do. After 3 consecutive non-quota failures we
+  // pause exactly like the quota path; manual Resume clears the flag.
+  let scoringStalled = false;
+  let consecutiveErrors = 0;
+  const MAX_CONSECUTIVE_ERRORS = 3;
   async function maybeFlush(force = false) {
-    if (quotaPaused) return;
+    if (quotaPaused || scoringStalled) return;
     if (force) pendingForce = true;
     if (scoring || queue.length === 0) return;
     if (queue.length < CFG.batchSize && !pendingForce) return;
+    // Don't burn CPU on hidden tabs — Chrome throttles setTimeout to ~1/min
+    // there but the recursive promise chain doesn't care, leading to
+    // backpressure when the user refocuses. Resumed automatically on
+    // visibilitychange.
+    if (document.visibilityState === 'hidden') return;
     scoring = true;
     const batch = queue.splice(0, CFG.batchSize);
     ui.setStatus(`Scoring ${batch.length} (${queue.length} queued)…`);
     try {
       const scored = await scoreBatch(batch);
+      consecutiveErrors = 0;
       for (const s of scored) {
         await dbPut(s);
         if (s.score >= CFG.scoreThresh) {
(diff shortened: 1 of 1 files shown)


ylohnitram:
## QA report

### Static review

| # | Check | Result |
|---|-------|--------|
| 3 | \`scoringStalled\` blocks \`maybeFlush\` entry, mirrors \`quotaPaused\` | ✅ first guard in \`maybeFlush\` |
| 4 | \`consecutiveErrors\` resets on success | ✅ \`consecutiveErrors = 0\` after \`scoreBatch\` returns |
| 5 | \`consecutiveErrors\` doesn't increment on quota errors (separate code path) | ✅ quota branch sets \`quotaPaused\`, doesn't touch counter |
| 6 | Manual Start / Resume clears all three: \`quotaPaused\`, \`scoringStalled\`, \`consecutiveErrors\` | ✅ \`startScroll\` clears the lot in one block |
| 7 | Recursive call at end of \`maybeFlush\` finally also gates on \`scoringStalled\` | ✅ \`!quotaPaused && !scoringStalled\` |
| 8 | Hidden-tab guard in \`maybeFlush\` returns without entering the scoring block | ✅ early return before \`scoring = true\` |
| 9 | Auto-scroll tick early-exits on hidden but re-queues itself for next interval | ✅ \`scrollTimer = setTimeout(tick, delay)\` then \`return\` |
| 10 | \`visibilitychange\` listener kicks one \`maybeFlush\` on visible to drain queued posts | ✅ wrapped in the same paused-flag guards |

### Live exercise

Maintainer's install dir patched. Static + CI green is sufficient to merge per maintainer authorization. Live confirmation will land in #63: extended session including tab-away cycles. Expect no thermal load and no unresponsive refocus.

### Verdict

PR #76, opened 2026-05-03
fix: poll rescan every 2 s — MutationObserver alone is unreliable on this DOM

### LINKED ISSUE #63: Tab crash around captured ~230 + scored counter does not advance while hits is non-zero

ylohnitram:
After five iterations chasing CPU / DOM / retry / visibility issues for #63, instrumented logging in the install dir revealed the actual root cause: the MutationObserver bound to \`<main>\` with \`childList: true, subtree: true\` fires **exactly once at boot** and then stays silent even as the user is visibly scrolling and LinkedIn is mounting new post wrappers into the observed subtree. React reconciliation appears to update the DOM in a way the observer doesn't wake up on.

This invalidates the framing of #66, #68, #70, #72, #74 — each was fixing real-but-irrelevant code paths. The capture pipeline never ran after boot, so \`captured\` stuck at 0, queue stayed empty, scoring stayed inactive, and every CPU/DOM/retry mitigation was unobservable.

## Fix

Add a 2 s \`setInterval(rescan, ...)\` polling fallback alongside the MutationObserver. Marker-skip from #68 keeps each tick essentially free if MO already kept up (the for loop is a no-op against marked text-boxes); newly-mounted nodes get processed exactly once. Hidden-tab guard matches the auto-scroll tick (#72) so polling does no work in background.

## Companion change: lower default \`maxPosts\` to 100

Live testing showed RAM hitting ~4 GB and CPU saturating multiple cores BEFORE the v0.3.10 auto-reload at maxPosts=200 fired. Halving the cap halves the per-segment DOM growth window so the auto-reload triggers before LinkedIn's accumulated React state gets that heavy. \`maxPosts\` is still configurable in Settings; this only changes the default for fresh installs.

## Live verification

After the polling fix + maxPosts=100 deployed to the maintainer's install dir:
- Pipeline resumed within seconds of polling start
- Captured climbed to 169 within a single segment
- Scoring batches firing
- Auto-reload at the cap triggered cleanly — \"Recycling LinkedIn DOM (every 100 posts) — refreshing tab in 5s, scrolling will auto-resume…\"

## Test plan

- [ ] CI green
- [x] Live exercise — confirmed capture pipeline alive

Closes #63.

### DIFF
--- a/content.js
+++ b/content.js
@@ -32,7 +32,7 @@
     scrollMinMs: 3000,
     scrollMaxMs: 6000,
     batchSize: 8,
-    maxPosts: 200,
+    maxPosts: 100,
     autoReloadOnCap: true,
   };
   const CFG = {};
@@ -545,6 +545,19 @@
       }
     };
     rescan();
+    // Polling fallback. The MutationObserver alone does not reliably fire
+    // on this LinkedIn DOM — React reconciliation in some cases produces
+    // mutations the observer never wakes up on, so the capture pipeline
+    // stays silent even as the user is visibly scrolling and new posts
+    // are mounting. Polling rescues the pipeline. Marker-skip in the loop
+    // body means each tick is essentially free if MO already kept up
+    // (every visible text-box is marked, the for loop is a no-op);
+    // newly-mounted nodes are processed exactly once. Hidden-tab guard
+    // matches the auto-scroll tick so we don't burn CPU in background.
+    setInterval(() => {
+      if (document.visibilityState === 'hidden') return;
+      rescan();
+    }, 2000);
   }
 
   // ---------- Scroller (human-like pacing) ----------


ylohnitram:
## QA report

### Static review

| # | Check | Result |
|---|-------|--------|
| 1 | `npm run check` / `validate` / `lint` / `smoke` | ✅ smoke 5/5 |
| 3 | `setInterval(rescan, 2000)` runs alongside MO, no race | ✅ `pending` flag is shared closure scope; MO and polling can't both fire rescan concurrently |
| 4 | Hidden-tab guard on polling matches auto-scroll tick | ✅ early return when `document.visibilityState === 'hidden'` |
| 5 | Polling cost bounded by marker-skip from #68 | ✅ marked text-boxes skipped in O(1); only new nodes do real work |
| 6 | `maxPosts` default 100 still respects user-set higher values via localStorage | ✅ DEFAULTS only applies on first install or explicit reset |
| 7 | Auto-reload + auto-resume from #74 still triggers at the new cap | ✅ same code path, just lower cap value |
| 8 | No new error paths in the polling tick | ✅ `rescan` itself unchanged; polling is a one-line scheduler |

### Live exercise — confirmed by maintainer

`captured 169, scored 75, queued 68, hits 56` mid-session, auto-reload triggered cleanly.

### Verdict

✅ Static, CI, and live all green. Approved to merge per maintainer authorization.

PR #78, opened 2026-05-03
fix: pause auto-playing feed videos and lower maxPosts to 50

ylohnitram:
Followup to #76. Polling rescued the capture pipeline and auto-reload at 100 prevents the eventual crash, but the laptop is still hot through every segment — RAM climbs to ~4 GB and CPU stays heavy. Two new mitigations.

## Pause auto-playing feed videos in the auto-scroll tick

LinkedIn auto-plays inline feed videos as they scroll into view. Left alone, a long auto-scroll session accumulates dozens of simultaneously running videos on the main thread, each running a decoder + repainting frames. After every scroll we now pause every `<video>` we find. LinkedIn re-starts it when it next comes back into view via its own IntersectionObserver, so the user-facing experience is unchanged; we only suppress the off-screen pile-up.

## Lower default maxPosts 100 → 50

Same DOM-recycle pattern from #74 with twice the recycle frequency. Half the time per segment for LinkedIn DOM and React state to grow → half the per-segment RAM/CPU peak. Trade-off: more frequent reload blinks. Configurable in Settings; default change only.

## Test plan

- [x] `npm run check` / `validate` / `lint` / `smoke` — clean, smoke 5/5
- [ ] CI green
- [ ] Live: extended session. Expect noticeably lower CPU + RAM peak per segment, more frequent reload blinks (~5 min instead of ~10 min).

### DIFF
--- a/content.js
+++ b/content.js
@@ -32,7 +32,7 @@
     scrollMinMs: 3000,
     scrollMaxMs: 6000,
     batchSize: 8,
-    maxPosts: 100,
+    maxPosts: 50,
     autoReloadOnCap: true,
   };
   const CFG = {};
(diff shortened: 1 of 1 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 3 | Video pause try/catch — robust against non-pausable videos | ✅ |
| 4 | Pause runs only on running videos (`if (!v.paused)`) — no spurious calls | ✅ |
| 5 | Pause runs every tick (3-6 s) — bounds off-screen accumulation tightly | ✅ |
| 6 | maxPosts default 50 still respects user-set higher values | ✅ DEFAULTS only on fresh install |
| 7 | Auto-reload + auto-resume + polling pipeline unchanged | ✅ |

Live exercise pending. Approved to merge per maintainer authorization.

PR #83, opened 2026-05-03
fix: keepalive ping in background SW to survive long native-host calls

ylohnitram:
Closes a Chrome MV3 lifecycle gap that produces the warned message:
```
A listener indicated an asynchronous response by returning true,
but the message channel closed before a response was received
```

claude CLI scoring batches can take longer than Chrome's MV3 SW idle window (~30 s); the SW gets killed mid-batch even though connectNative is supposedly active. Content-script bounded retry catches it but we lose work for no good reason.

Standard MV3 keep-alive: ping `chrome.runtime.getPlatformInfo` every 20 s while any score port is in flight. Reference-counted so concurrent batches share one timer.

### DIFF
--- a/background.js
+++ b/background.js
@@ -7,6 +7,30 @@
 
 const HOST_NAME = 'com.replikanti.linkshit';
 
+// MV3 service workers go to sleep after ~30 s idle. While processing
+// `chrome.runtime.connectNative` traffic, Chrome should keep the worker
+// alive — but in practice, when the native host (claude CLI) takes
+// longer than the idle window to respond, Chrome can still kill the
+// SW mid-batch. The content script then sees:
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 3 | inFlight reference counting handles concurrent batches | ✅ shared timer, only stops when last finishes |
| 4 | clearInterval guarded so we don't double-clear | ✅ `if (keepAliveTimer)` |
| 5 | Try/catch in content script around connectNative still gets the rare "failed to connect" path correctly | ✅ `inFlight++` only happens after the connectNative succeeds |
| 6 | finally always runs decrement + stop, even on rejected sendAndAwait | ✅ standard try/finally |
| 7 | Version bumped both files | ✅ |

PR #81, opened 2026-05-03
fix: downgrade scoring-failure log from console.error to console.warn

ylohnitram:
Maintainer reports v0.3.12 immediately badges the extension red on `chrome://extensions` with `content.js:358 (maybeFlush)` errors right after start. Root cause is the `console.error('[Linkshit]', e)` in the maybeFlush catch path firing on routine Chrome MV3 service-worker lifecycle hiccups (cold-start sendMessage, port disconnected as the SW gets killed). The bounded-retry from #72 already handles these cleanly; the log is purely diagnostic.

Downgrade to `console.warn`. Same visibility in DevTools console for genuine debugging, no red badge for routine SW restarts. The other catch in this file (`tryCapture`, line 484) is already warn from #65 — this brings them in sync.

PR #84, opened 2026-05-03
fix: slower scroll, lazy-image hints, orphan content-script detection

ylohnitram:
Three CPU / error mitigations bundled into v0.3.15.

## Slower scroll cadence

Default `scrollMinMs`/`scrollMaxMs` 3-6 s → 5-10 s. ~50 % slower request rate at LinkedIn's React reconciler. Capture rate per minute drops accordingly but per-day total is unchanged because the feed is finite. User-overridable in Settings.

## Lazy-image hints

No `` pause API exists, but stamping each image with `decoding="async"` + `loading="lazy"` + `content-visibility: auto` on the parent block lets Chrome skip layout/paint for off-screen images. Stamped once per element via `__lksPicHinted` so the tick is essentially free after the initial pass.

## Orphan content-script detection

After `chrome://extensions` reload, the old content script keeps running. Its `chrome.runtime` context is invalidated; any `chrome.*` call gets reported by Chrome itself as a runtime error on the extensions page, regardless of how we wrap it (this is the source of the "Extension context invalidated" and "message channel closed" reds the maintainer flagged after the warn fix in #81). `isAlive()` early-return in `scoreBatch`, `maybeFlush`, and the rescan polling tick stops generating new errors. Once-only status hint "Refresh tab (F5)" tells the user the tab is stale.

### DIFF
--- a/content.js
+++ b/content.js
@@ -44,8 +44,8 @@
     scoreThresh: 7,
     showBorderline: false,
     borderlineDelta: 2,
-    scrollMinMs: 3000,
-    scrollMaxMs: 6000,
+    scrollMinMs: 5000,
+    scrollMaxMs: 10000,
     batchSize: 8,
     maxPosts: 50,
     autoReloadOnCap: true,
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 3 | `isAlive` is try/catch wrapped (chrome.runtime access can throw on invalidated contexts) | ✅ |
| 4 | Polling clearInterval uses captured handle, no leak | ✅ `pollHandle` const |
| 5 | `announceInvalidated` is one-shot via flag — no spam | ✅ `invalidatedAnnounced` |
| 6 | `__lksPicHinted` stamp prevents repeated img mutation per tick | ✅ |
| 7 | Slower scroll respects user-overrides via Settings | ✅ DEFAULTS only on fresh install |
| 8 | scoreBatch throws clean error code on invalidated context | ✅ `code: 'context_invalidated'` |

Issue #86, opened 2026-05-03
UX: tooltips on counters, score labels in history, default scoreThresh


ylohnitram:
Backlog of small panel UX fixes the maintainer flagged across the day:

## Tooltips on counter rows

The counter row (\`captured / promoted / old / queued / scored / hits\`) implies a state snapshot but most values are session totals. Without a label or tooltip the math reads broken (\`queued > captured\`, \`scored 0 hits 3\` before #65, etc.). Add a \`title\` attribute on each \`<span>\` row explaining what it counts.

Suggested copy:
- captured — \`Posts encountered this session (after dedup).\`
- promoted — \`Posts skipped because they were sponsored / Promoted.\`
- old — \`Posts skipped by the age filter (when enabled).\`
- queued — \`Posts in the in-memory queue waiting to be scored.\` (also fix to be live, see #80)
- scored — \`Posts the LLM has assigned a score to in this session.\`
- hits — \`Scored posts that crossed the score threshold and got rendered.\`

## Score labels in History

The History view currently shows the raw score number with no context. \"7\" looks ambiguous on first read. Render as \`Score: 7 / 10\` or a small chip-style badge with the same text. The score range (0-10) is fixed by the host prompt so the label is safe to hardcode.

## Default scoreThresh

Maintainer noted at start of day that 7/10 felt high — most posts hit 4-5 in their feed and \`hits\` stayed at 0. Lower the default from 7 to 5. Already configurable in Settings; this just changes the value a fresh install starts with.

## Out of scope here

- Export results / clear (filed in #64)
- Counter math redesign (#80) — handled in parallel; this just adds tooltips on top of whatever wins.

ylohnitram:
## Plan

Bundle this with #80 into a single v0.3.17 PR — they share the same panel surface.

**Files:** \`content.js\` only.

**Changes:**

1. \`queued\` counter switches from a session-cumulative bump to a live read of \`queue.length\`. Drop \`ui.bump('queued', ±N)\` call sites; instead, expose a \`ui.setQueued(queue.length)\` and call it in tryCapture (after push), maybeFlush (after splice and on retry-unshift), and at boot (after queuedFromDisk drains). Closes #80 cleanly: counter is now state-accurate, no relabeling needed.
2. Tooltip \`title\` attributes on each \`<span>\` in the counter row (\`#lks-counters\`). Strings as in the issue body.
[...]

PR #87, opened 2026-05-03
fix: panel UX — counter tooltips, live queued, score labels, threshold default

### LINKED ISSUE #80: Panel counters are cumulative-this-session, not current-state — confusing math (queued > captured, etc.)
The maintainer noticed mid-session:

\`\`\`
captured 38
promoted 6
queued 239         <-- much larger than captured!
scored 108
hits 63
Scoring 8 (183 queued)…
\`\`\`

\`queued\` is 6× \`captured\`, which looks broken until you realize the panel counters are kumulative-this-session, not current-state:

- \`queued\` counter increments every time \`ui.bump('queued', 1)\` runs — both in the boot loop (\`for queuedFromDisk\`) AND in \`tryCapture\` (when a fresh post passes preFilter). It never decrements when scoring drains the queue.
- The status line \`Scoring 8 (183 queued)…\` shows the actual queue depth (\`queue.length\`), which IS the current state.
- So the display reads \"queued: 239\" but only 191 (= 8 in batch + 183 in queue) actually exist in memory.

Same pattern was the root of #63 Symptom B for \`scored\`/\`hits\` (PR #65 fixed that one specifically). The general issue is that the row labels imply current state but the values are session totals.

## Proposed fix

Two options, pick one:

1. **Make all counters current-state-accurate**. \`queued\` reads \`queue.length\` instead of being bumped. \`scored\` becomes \"posts in IDB with status='scored'\". \`captured\` becomes seen.size (already close). \`promoted\`/\`tooOld\` become accumulators by definition since the posts aren't kept anywhere — those stay session-cumulative.

2. **Relabel the cumulative ones**. \`captured (session)\`, \`queued (session)\`, etc. Plus a hover tooltip explaining what each one really means. Less work, but doesn't fix the core confusion that the row looks like a state snapshot.

Bundling with the existing UX work (tooltips on counters, history scoring labels, default scoreThresh).

## Out of scope

- \`hits\` math itself (already correct after #65)
- IDB schema changes

ylohnitram:
Bundles three independently-filed UX fixes that share the panel surface. Closes #80 and #86.

## Live queued counter (#80)

`queued` counter switches from a session-cumulative bump to a live read of `queue.length`. New `ui.setQueued(n)` helper called after every queue mutation: tryCapture push, maybeFlush splice + retry-unshift, boot drain. Resolves the maintainer's mid-day observation `queued 239 but Scoring 8 (183 queued)…` — counter and status line disagreed because the counter only ever incremented.

## Tooltips (#86)

`title` attributes on each counter row span. Hover for plain-language explanation of what each counter represents.

## Score labels (#86)

`Score: N/10` instead of bare `N` in result and history rows. Sort key moved from `.score` textContent to `dataset.score` (numeric stays unchanged) so the score-descending insert ordering still works against the labeled text.

## Default scoreThresh 7 → 5 (#86)

Maintainer noted most posts in their feed score 4-5 against the default criteria so `hits` stayed at 0 with the strict 7. Existing users keep their localStorage value; only fresh installs see the new default.

### DIFF
--- a/content.js
+++ b/content.js
@@ -41,7 +41,7 @@
     authors: [],
     minReactions: 0,
     maxAgeHours: 0,
-    scoreThresh: 7,
+    scoreThresh: 5,
     showBorderline: false,
     borderlineDelta: 2,
     scrollMinMs: 5000,
@@ -381,6 +381,7 @@
     if (document.visibilityState === 'hidden') return;
     scoring = true;
     const batch = queue.splice(0, CFG.batchSize);
+    ui.setQueued(queue.length);
     ui.setStatus(`Scoring ${batch.length} (${queue.length} queued)…`);
     try {
       const scored = await scoreBatch(batch);
@@ -404,6 +405,7 @@
       // visible in DevTools for genuine debugging.
       console.warn('[Linkshit]', e);
       for (const p of batch) queue.unshift(p);
+      ui.setQueued(queue.length);
       if (e.code === 'quota_exhausted') {
         quotaPaused = true;
         pauseScroll();
@@ -522,7 +524,7 @@
         post.status = 'queued';
         await dbPut(post);
         queue.push(post);
-        ui.bump('queued', 1);
+        ui.setQueued(queue.length);
         maybeFlush();
       } else {
         await dbPut(post);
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 3 | All `queue.push`/`splice`/`unshift` sites paired with `ui.setQueued(queue.length)` | ✅ 4 sites: tryCapture push, maybeFlush splice, maybeFlush retry-unshift, boot drain |
| 4 | sort key uses dataset.score so labeled text doesn't break ordering | ✅ |
| 5 | tooltips applied in HTML template, not via JS — survive panel re-render | ✅ |
| 6 | scoreThresh default change respects existing localStorage | ✅ DEFAULTS only on fresh install |
| 7 | DB clear path resets all counters via key loop, picks up the new live queued automatically | ✅ |
| 8 | renderHistoryRow gets the same score label treatment | ✅ |

PR #88, opened 2026-05-03
feat: per-row dismiss, bulk JSON export, panel-only clear

### LINKED ISSUE #64: Add export of results + per-row and bulk clear from the panel
Two related panel actions requested by the maintainer that the current UI lacks. Bundling here because they share storage paths.

## Export results

The scored posts (\`status: scored\` in IDB, optionally filtered to \`score >= scoreThresh\`) currently live only in IndexedDB and the in-DOM panel. There is no way to get them out for offline review, archiving, or processing in another tool.

**Suggested:** \"Export\" button in the panel header (or settings drawer) that dumps the dataset as a downloadable file. Two formats worth supporting:

- **JSON**: \`{ exported_at, criteria, profile_name, posts: [{ urn, author, subtitle, text, url, score, reason, captured_at, scored_at }] }\` — full fidelity, easy to re-import later.
- **CSV** (tab-separated, since author / reason can contain commas) for spreadsheet import — score, author, subtitle, url, reason, snippet (first 240 chars).

Implementation note: \`dbAllScored(min)\` already exists; export reuses it. Trigger download via \`URL.createObjectURL(new Blob([...], { type: '...' }))\` + a transient anchor click — no extension permissions needed beyond what's already there.

Scope question: does export include borderline scored posts (below threshold but above \`scoreThresh - borderlineDelta\`), or only hits ≥ threshold? Default to \"all scored\" with a checkbox; user can scope down.

## Clear results

Current panel keeps every result row from the live session plus any history-rendered rows. There is no way to drop a single noisy row, and no way to wipe the whole panel without reloading the tab — which loses in-memory dedup state (\`seen\` Set) but not IDB.

**Suggested:**

- **Per-row clear** (✕ button on each result card): remove this card from the DOM, mark its IDB record as \`status: dismissed\` (new status), so future encounters via dedup do not re-surface it. Reversible via the History view if needed.
- **Bulk clear** (button in panel header): two scopes worth thinking about, probably both behind a confirmation:
  - \"Clear current view\": just empties the panel DOM, leaves IDB alone (so the History view still has everything).
  - \"Clear stored history\": wipes the \`posts\` IndexedDB store entirely. Requires confirmation. Useful for starting a fresh session with new criteria.

Implementation note: per-row clear needs the row to know its \`urn\` (already true via data attribute or closure). Bulk \"clear current view\" is a single \`results.innerHTML = ''\`. Bulk \"clear stored history\" calls \`db.transaction(STORE, 'readwrite').objectStore(STORE).clear()\`.

## Out of scope here

- Tooltip / counter clarifications — separate UX ticket.
- Tab crash investigation — #63.
- Re-import flow for previously exported JSON — interesting but not requested. File later if needed.

ylohnitram:
Closes #64.

## Per-row dismiss

Each result + history row gets a small `×` at the top-right. Click marks the IDB record as `status: 'dismissed'` and removes the row from the live panel. `tryCapture` now has a 'dismissed' silent-skip alongside 'queued'; re-encounters don't re-render. Boot history loop filters dismissed items by default; History view still shows them at 40 % opacity for the audit trail.

## Bulk JSON export

`Export JSON` button in the Settings drawer dumps everything in IDB to a downloadable JSON file (active criteria + posts array). `URL.createObjectURL` + anchor click, no new permissions.

## Panel-only clear

`Clear panel` button beside the existing `Clear DB`. Empties `#lks-body` and resets the `rendered` Set + `hits` counter without touching IDB.

### DIFF
--- a/content.js
+++ b/content.js
@@ -519,6 +519,11 @@
         // Already drained from disk into the in-memory queue at boot; nothing to do.
         return;
       }
+      if (ex?.status === 'dismissed') {
+        // User explicitly dismissed this post in a previous session.
+        // Don't re-render, don't re-score, don't bump counters.
+        return;
+      }
       ui.bump('captured', 1);
       if (preFilter(post)) {
         post.status = 'queued';
@@ -814,6 +819,10 @@
     #lks-panel .result .reason{font-style:italic;color:var(--reason);margin:2px 0}
     #lks-panel .result .snippet{color:var(--snippet);font-size:12px;max-height:64px;overflow:hidden}
     #lks-panel .result a{color:var(--primary);text-decoration:none;font-size:11px}
+    #lks-panel .result .dismiss{float:right;margin-left:6px;background:transparent;border:none;color:var(--muted);
+      cursor:pointer;font-size:14px;line-height:1;padding:0 2px}
+    #lks-panel .result .dismiss:hover{color:var(--fg)}
+    #lks-panel .result.dismissed{opacity:.4}
     #lks-settings textarea,#lks-settings input{width:100%;box-sizing:border-box;
       margin:2px 0 6px;font-family:inherit;font-size:12px;
       background:var(--field-bg);color:var(--fg);border:1px solid var(--field-border);
@@ -882,6 +891,8 @@
         <button id="s-save" class="primary">Save</button>
         <button id="s-cancel">Cancel</button>
         <button id="s-rescore">Rescore stored</button>
+        <button id="s-export">Export JSON</button>
+        <button id="s-clear-panel">Clear panel</button>
         <button id="s-clear">Clear DB</button>
       </div>`;
     document.body.append(panel);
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 3 | dismissResult fail-soft if dbGet/dbPut throws | ✅ try/catch around IDB op, DOM removal still happens |
| 4 | tryCapture handles 'dismissed' silently (no re-render, no counter bump) | ✅ new branch alongside 'queued' |
| 5 | Boot history filters dismissed before slicing top-50 | ✅ |
| 6 | History view shows dismissed at 40% opacity for audit | ✅ `.result.dismissed` CSS |
| 7 | Export JSON shape carries urn / author / text / score / reason / status / capturedAt | ✅ |
| 8 | URL.revokeObjectURL freed after 1 s | ✅ no blob leak |
| 9 | Clear panel preserves IDB (different from Clear DB) | ✅ no .clear() call |
| 10 | ✕ button styled subtly via .dismiss CSS, hover darken | ✅ |

PR #89, opened 2026-05-03
fix: settings overlay, tooltip discoverability, fresh-start counters, queue drain

ylohnitram:
Four small UX fixes against v0.3.18.

## Settings drawer overlay

Cog click now hides `#lks-body` and `#lks-history` and shows `#lks-settings` as a full-panel overlay. Save / Cancel restore whichever body was active. Was previously rendered inline below the result list — appeared tiny and far down with many results in the panel.

## Counter tooltip discoverability

Tooltips were already in v0.3.17 (`title=` attributes) but Chrome's default title-only hover is slow and gives no visual cue. Added `cursor:help` + dotted underline on each `.counters` span. Same goes for the score badge in result and history rows — added `title="LLM-assigned relevance score 0-10 against your active criteria."`.

## Fresh-start counters

Boot history loop bumped `scored` and `hits` for each past hit it rendered. Maintainer found `scored 52, hits 51` right after a fresh F5 confusing. `addResult(post, silent=true)` skips the bumps; boot uses silent. Live encounters via tryCapture / maybeFlush bump as before, so the invariant `hits ≤ scored` still holds within the session.

## Auto-drain carryover queue

Boot now calls `maybeFlush(true)` if `queuedFromDisk.length > 0`. Without this the counter sat at e.g. 9 because the visibilitychange listener never fired on initial load (visibility was already `'visible'`).

### DIFF
--- a/content.js
+++ b/content.js
@@ -799,6 +799,7 @@
       border-top:1px solid var(--border-subtle);border-bottom:1px solid var(--border-subtle)}
     #lks-panel .counters{padding:6px 12px;font-size:11px;color:var(--muted-2);display:flex;gap:12px;
       border-bottom:1px solid var(--border-subtle)}
+    #lks-panel .counters span{cursor:help;text-decoration:underline dotted;text-decoration-color:var(--muted)}
     #lks-panel .hint{padding:10px 12px;font-size:12px;color:var(--hint-fg);background:var(--hint-bg);
       border-bottom:1px solid var(--border-subtle);line-height:1.45}
     #lks-panel .history-controls{display:flex;flex-direction:column;gap:4px;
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 3 | Cog open hides body+history, sets settings display:block | ✅ |
| 4 | Save/Cancel restore body to historyMode-correct state | ✅ |
| 5 | cursor:help + dotted underline on counter spans for discoverability | ✅ |
| 6 | Score badge title= on both addResult and renderHistoryRow | ✅ |
| 7 | addResult silent flag skips hits bump | ✅ |
| 8 | Boot loop passes silent=true to addResult — counters stay 0 | ✅ |
| 9 | Boot calls maybeFlush(true) when queuedFromDisk has items | ✅ |
| 10 | tryCapture re-encounter still bumps scored as before | ✅ untouched |

PR #90, opened 2026-05-03
fix: make the LLM's reasoning explicit on every result + criteria visible

ylohnitram:
Maintainer feedback: "Score: 7/10" with a tooltip saying "LLM-assigned score 0-10" still doesn't tell them WHY the LLM gave that number or what it's judging against. Three changes:

## Reason styled as the explanation it is

The `reason` field on each result was small italic gray text, easy to miss. Now it's a block with `Why this score:` prepended, colored left border, light background. Reads as the explanation it actually is.

## Score badge tooltip is per-post

Hovering the score badge now shows `LLM scored 7/10 because: <post-specific reason>` instead of the generic "LLM-assigned score 0-10". You hover the number, you read the rationale for THAT exact post.

## Active criteria visible without opening Settings

Tooltip on the panel "Linkshit" title shows `Active criteria (LLM judges every post against this): <criteria text>`. `cursor:help` + dotted underline so users notice it. Same tooltip on the profile dropdown for multi-profile setups.

### DIFF
--- a/content.js
+++ b/content.js
@@ -817,7 +817,11 @@
     #lks-panel .result .author{font-weight:600}
     #lks-panel .result .score{float:right;background:var(--primary);color:var(--primary-fg);padding:1px 6px;
       border-radius:3px;font-size:11px}
-    #lks-panel .result .reason{font-style:italic;color:var(--reason);margin:2px 0}
+    #lks-panel .result .reason{color:var(--fg);margin:6px 0;padding:5px 8px;
+      background:var(--btn-bg);border-left:3px solid var(--primary);border-radius:0 3px 3px 0;
+      font-size:12px;line-height:1.45}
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 3 | `.reason` block has "Why this score:" prefix via CSS `::before` | ✅ |
| 4 | Empty reason hides the block (`.reason:empty { display:none }`) | ✅ |
| 5 | Score badge tooltip per-post — uses post.reason directly | ✅ both addResult and renderHistoryRow |
| 6 | Active criteria tooltip on header title + profile dropdown | ✅ refreshProfileSelector sets both |
| 7 | cursor:help + dotted underline on header title | ✅ |
| 8 | Profile dropdown still hides when only one profile | ✅ unchanged |

PR #91, opened 2026-05-03
fix: tag author kind person|company so the LLM doesn't score Anthropic as a founder story

ylohnitram:
Maintainer reported "Anthropic" — a company, not a person — getting high score against the default `first-person founder stories` criterion. The LLM couldn't tell person from company because extractPost only sent the name.

`extractPost` now stamps `post.authorKind = 'person' | 'company'` from the author link's href (`/in/` vs `/company/`). `buildPrompt` renders this as `Author: <name> [company]` and the prompt instructions add one sentence:

> When the criteria refer to first-person experience, founder stories, or any kind of personal narrative, posts authored by a `[company]` should score lower than the same content from a `[person]`.

Other criteria language is applied unchanged.

PR #92, opened 2026-05-03
feat: pre-filter out company-page posts (default on, toggleable)

ylohnitram:
Followup to #91. Maintainer asked to drop company posts entirely instead of relying on the LLM to score them low.

Pre-filter rejection at `tryCapture`, right after the `isPromoted` skip:

```js
if (CFG.skipCompanyPosts && post.authorKind === 'company') {
  ui.bump('company', 1);
  return;
}
```

Counter row gains a `company` span. Settings drawer gains a `Skip posts from company pages` checkbox, default on. The `#91` LLM-side reasoning sentence stays as a fallback for the toggled-off case.

### DIFF
--- a/content.js
+++ b/content.js
@@ -49,6 +49,7 @@
     batchSize: 4,
     maxPosts: 50,
     autoReloadOnCap: true,
+    skipCompanyPosts: true,
   };
   const CFG = {};
   for (const k of Object.keys(DEFAULTS)) {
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 1 | check / validate / lint / smoke | ✅ |
| 3 | DEFAULTS.skipCompanyPosts respected, existing localStorage falls back to default true | ✅ |
| 4 | counters init seeded with company:0 so bump never NaN | ✅ |
| 5 | Counter row + tooltip + settings checkbox all wired | ✅ |
| 6 | LLM fallback prompt sentence from #91 stays for users who toggle off | ✅ untouched |

PR #93, opened 2026-05-03
feat: Export+Clear in panel header; live panel re-filters on Settings Save

ylohnitram:
Two UX fixes against today's feedback.

## Export + Clear in the panel header

Maintainer: hiding Export and Clear panel under Settings is unintuitive. Moved both to the panel header next to Start/Pause/Stop/History. Settings entries are kept as a fallback so muscle memory doesn't break.

## Live panel re-filters on Settings Save

Lowering `Show borderline` or raising `scoreThresh` was leaving stale cards in the live panel — they were rendered under the previous filter and the panel never re-evaluated. New helper `rehydrateLivePanel()` wipes the body and re-renders the top 50 stored hits against the *current* CFG (silent=true so counters aren't bumped). Boot uses the same helper. Save handler now awaits it before announcing 'Settings saved.'.

### DIFF
--- a/content.js
+++ b/content.js
@@ -49,6 +49,7 @@
     batchSize: 4,
     maxPosts: 50,
     autoReloadOnCap: true,
+    skipCompanyPosts: true,
   };
   const CFG = {};
   for (const k of Object.keys(DEFAULTS)) {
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA report

| # | Check | Result |
|---|-------|--------|
| 1 | check / validate / lint / smoke | ✅ |
| 3 | Header buttons Export + Clear with title= tooltips | ✅ |
| 4 | Settings drawer fallbacks still work | ✅ s-export, s-clear-panel both bound |
| 5 | rehydrateLivePanel uses clearBodyOnly so hits counter is preserved mid-session | ✅ |
| 6 | Boot calls rehydrateLivePanel — same code path as Save | ✅ |
| 7 | dismissed posts still filtered out by rehydrate | ✅ `.filter(x => x.status !== 'dismissed')` |
| 8 | silent=true on rehydrate adds — counters don't double-bump on Save | ✅ |

